PR #43, opened 2022-02-10
Adding commands to build and test scripts for React testing

### LINKED ISSUE #41: Update build.sh and test.sh with react test info
#31 
We need the build.sh and test.sh scripts to install NodeJS and npm, then install the node_modules, then run npm test in the correct folder

AshlynAdamson:
Closes #41
Commands needed for React code and React tests. 
Resources used to figure these out:
https://github.com/facebook/create-react-app/issues/11043
https://www.digitalocean.com/community/tutorials/how-to-install-node-js-on-ubuntu-20-04

### DIFF
--- a/.github/workflows/beima.yml
+++ b/.github/workflows/beima.yml
@@ -19,6 +19,6 @@ jobs:
 
       # Runs the build and test scripts
       - name: Build
-        run: build.sh
+        run: ../build.sh
       - name: Test
-        run: test.sh
+        run: ../test.sh
--- a/.github/workflows/beima.yml
+++ b/.github/workflows/beima.yml
@@ -19,6 +19,6 @@ jobs:
 
       # Runs the build and test scripts
       - name: Build
-        run: ../build.sh
+        run: ./build.sh
       - name: Test
-        run: ../test.sh
+        run: ./test.sh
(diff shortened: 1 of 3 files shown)


AshlynAdamson:
- [ ] get the darn workflow to run with my changes

.github/workflows/beima.yml
```diff
@@ -1,9 +1,8 @@
 name: beima
 
 on:
-  pull_request:
```
josephmoore99, line 4: I would think that we would want to keep this no?

AshlynAdamson: I think once we get every command we need, we can change it back. I put on push in here, so whenever a push happens on any branch it will run. It may be useful as we are adding stuff to it and working out the kinks

josephmoore99 (approved):

.github/workflows/beima.yml
AshlynAdamson: Having both the Push and PR trigger caused the workflow to run twice every time I would push to my branch that was in a PR. So I think for now the Push trigger will cover what we need. We can change it how to something a bit more specific once we get all the commands we need in the build and test scripts.

thessbsu (approved):

PR #44, opened 2022-02-10
Add Unit Testing

josephmoore99:
Set up the framework for adding unit tests to the code. I have set up some sample tests using different variations of "Assert" as examples for each of the classes. Right now they should all just pass no matter what.


Kenny-Miller (approved):

AshlynAdamson (approved):
Pulled down branch, built both projects, and ran the tests and they all passed. This setup is nice, having the source and tests in the same solution. It will make future development easier being able to have everything together to quickly verify changes still pass/implement new tests for new features

PR #55, opened 2022-02-11
Database schema creation

### LINKED ISSUE #23: Database Service - Handle Devices Only
#2 

To store information and metadata about the devices, we will be using MongoDB. To interact with our database, we need a service to perform some common CRUD operations. 

As a developer, I want to be able to use a database service that has abstracted out basic CRUD operations.

- [x] #24
- [x] #26
- [x] #27
- [x] #29
- [x] #33 
- [x] #60

roguedeleter:
Closes #23 

This is the proposed schema I came up with. The "fields" object in the device schema are the custom fields defined in the device template schema. The "customFieldUid" key will be a randomly generated UID.  This is to ensure that field names can be changed, without having to update all objects in the database.


Kenny-Miller (commented):
I would add a location too the building schema

BEIMA.DB.Schemas/building-schema
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
roguedeleter: Resolved

BEIMA.DB.Schemas/building-schema
```diff
@@ -0,0 +1,6 @@
...
+    "buildingId": String,
```
josephmoore99: Looking at the email, it looked like he also wanted to have the building number.

BEIMA.DB.Schemas/device-schema.txt
```diff
@@ -0,0 +1,25 @@
+{
+    "templateId": String,
```
AshlynAdamson, line 2: It seems like this field will refence a device template (device type) ID to determine the type of the device, but I think we also need a unique ID per device entry

BEIMA.DB.Schemas/device-schema.txt
```diff
@@ -0,0 +1,25 @@
...
+    "notes": String,
+    "location": {
+        "buildingId": String,
+        "notes": String,
+        "latitude": String,
+        "longitude": String
+    },
+    "lastModified": {
+        "date": Date,
+        "user": String
```
josephmoore99, line 18: Should we add in a user table? Maybe every time we have a login from a new user, we could add it into the user table. Might make for better tracking and could make it easier for use to assign certain access to specific users later on.

BEIMA.DB.Schemas/building-schema
roguedeleter: The buildingId field is a randomly generated UID to allow the ability to modify properties of the building without having to modify all documents in the database. The building number (or name) will be stored in the "name" field. Any other information can be stored in the "description" field.

BEIMA.DB.Schemas/device-schema.txt
AshlynAdamson: I think this will come later when we start work on the authentication part of BEIMA

BEIMA.DB.Schemas/device-schema.txt
roguedeleter: MongoDB will automatically populate a unique ObjectID whenever you insert into the database

BEIMA.DB.Schemas/building-schema
josephmoore99: I think that he wants number and name to be two separate entities though right? In the parameter column he lists it as "Building # & name"

BEIMA.DB.Schemas/device-schema.txt
roguedeleter: Yes, this is the ultimate intention, but not part of the scope of this task/PR. We will have a collection named "users", and there will be a record for each user containing login information and other information about the user.

This "user" field will store the username of the user who last modified the device.

BEIMA.DB.Schemas/building-schema
AshlynAdamson: We should change the 'name' field to be 'number' and 'description' to 'name' to make that connection a bit clearer.

BEIMA.DB.Schemas/device-template-schema.txt
```diff
@@ -0,0 +1,15 @@
+{
```
AshlynAdamson, line 1: Not a huge deal, but could we change the name of this file to be 'device-type-schema'? It took me longer to understand what template meant in this sense since aren't all schemas a template for future entries?

BEIMA.DB.Schemas/building-schema
roguedeleter: @josephmoore99 ah I see, yes the naming of the fields was definitely a bit confusing. I will use @AshlynAdamson's suggestion and change it to "name" and "number".

BEIMA.DB.Schemas/device-template-schema.txt
roguedeleter: Resolved

BEIMA.DB.Schemas/building-schema
roguedeleter: This has been fixed

josephmoore99 (approved):

Kenny-Miller (approved):

thessbsu (approved):

PR #57, opened 2022-02-12
Setup ft framework

### LINKED ISSUE #53: Set up initial FT framework
References user story #35

As a developer, when implementing a feature, then there should be framework in which functional tests can be added to in order to test requests against the backend API from an external application. This can likely be done with a new NUnit project that can be run against a running instance of the complete backend application.

josephmoore99:
Closes #53

Implemented the backend FT framework. Included is an http client class to allow us to interface with the running backend instance. Additionally I have included a sample functional test class.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -19,7 +19,7 @@ sudo apt install nodejs || { echo "Installing NodeJS failed"; exit 1; }
 # Install npm
 sudo apt install npm || { echo "Installing npm failed"; exit 1; }
 # Install Azure functions core tools
-npm i -g azure-functions-core-tools@4
+npm i -g azure-functions-core-tools@4 || { echo "Installing Azure functions core tools failed"; exit 1; }
 # Change directory to BEIMA.Client
 cd ./BEIMA.Client || { echo "Changing directories failed"; exit 1; }
 # Install node_modules
--- a/build.sh
+++ b/build.sh
@@ -18,6 +18,8 @@ sudo apt install microsoft-edge-dev || { echo "Install edge failed"; exit 1; }
 sudo apt install nodejs || { echo "Installing NodeJS failed"; exit 1; }
 # Install npm
 sudo apt install npm || { echo "Installing npm failed"; exit 1; }
+# Install Azure functions core tools
+npm i -g azure-functions-core-tools@4
 # Change directory to BEIMA.Client
 cd ./BEIMA.Client || { echo "Changing directories failed"; exit 1; }
 # Install node_modules
--- a/test.sh
+++ b/test.sh
@@ -1,18 +1,15 @@
 #!/bin/bash
-run_backend_locally(){
-    echo "STARTING LOCAL API"
-    func start
-}
-
 # Change directory to BEIMA.Backend.Test
 cd ./BEIMA.Backend.Test
 # Run backend unit tests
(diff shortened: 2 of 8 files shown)


test.sh
```diff
@@ -3,10 +3,22 @@
 cd ./BEIMA.Backend.Test
 # Run backend unit tests
 dotnet test ./BEIMA.Backend.Test.csproj || { echo "Unit tests failed"; exit 1; }
-# Change back to root directory
+# Navigate to backend project
+cd ../
+cd ./BEIMA.Backend
+# Start the backend API locally
+func start BEIMA.Backend.csproj --csharp &
```
josephmoore99, line 10: The '&' at the end allows us to start the API in the background without having to wait for it to exit to continue on to the rest of the code.

test.sh
```diff
@@ -3,10 +3,22 @@
...
+# Switch to FT directory
+cd ../
+cd ./BEIMA.Backend.FT
+# Give enough time for local API to finish starting up
+sleep 15
```
josephmoore99, line 15: Bit hacky, but couldn't think of anything better. open to ideas.

build.sh
```diff
@@ -18,6 +18,8 @@ sudo apt install microsoft-edge-dev || { echo "Install edge failed"; exit 1; }
 sudo apt install nodejs || { echo "Installing NodeJS failed"; exit 1; }
 # Install npm
 sudo apt install npm || { echo "Installing npm failed"; exit 1; }
+# Install Azure functions core tools
+npm i -g azure-functions-core-tools@4
```
AshlynAdamson, line 22: We should have a failing message with an exit code here as well

test.sh
AshlynAdamson: When I was working on the cypress testing, I also needed to start the front end server and test. Some of the solutions I found across the web included having a sleep command to wait for the server. I was able to find an npm package that started the server and then ran tests, if there's something like that out there to use on our backend, that would be cool. But for now I think this is fine

test.sh
```diff
@@ -3,10 +3,22 @@
...
 # Run backend unit tests
 dotnet test ./BEIMA.Backend.Test.csproj || { echo "Unit tests failed"; exit 1; }
-# Change back to root directory
+# Navigate to backend project
+cd ../
+cd ./BEIMA.Backend
+# Start the backend API locally
+func start BEIMA.Backend.csproj --csharp &
+# Switch to FT directory
+cd ../
```
AshlynAdamson, line 12: I'm running the backend tests on my machine, and I found if you cd into BEIMA.Backend and run `dotnet test` from there it will run both of our backend test projects. We could use that instead and save a few lines of code here

AshlynAdamson (approved):

Kenny-Miller (approved):

test.sh
thessbsu: I've played around with it a bit. When I run it with no wait it eventually fails with a connection refused message.
with the sleep 20 it seems to work ok

with this it fails later (for me), but it isn't related to the project not running:

until pids=$(pidof func start BEIMA.Backend.csproj --csharp &)
do
    sleep 1
done

this is from the first answer when searching for "sh wait for process to start"
There's other results, but all seem to use a loop to wait for the PID or a log message of some type

PR #67, opened 2022-02-15
Database Service - Handle Devices Only

### LINKED ISSUE #23: Database Service - Handle Devices Only

### LINKED ISSUE #24: DB Service - Insert Device
#23 

As a developer, I should have a service that can insert a device, given a template ID and the required metadata.

### LINKED ISSUE #26: DB Service - Update Device
#23 

As a developer, I should have a service that can update a device, given a unique device ID and the required metadata.

### LINKED ISSUE #27: DB Service - Get Device
#23 

As a developer, I should have a service that can return information about a device, given a device ID.

### LINKED ISSUE #29: DB Service - Delete Device
#23 

As a developer, I should have a service that can delete a device, given a device ID.

### LINKED ISSUE #60: Create unit tests for MongoConnector

roguedeleter:
Closes #23 - story
Closes #24 
Closes #26
Closes #27 
Closes #29 
Closes #60

This PR implements the MongoConnector. It abstracts out the basic CRUD operations we need for the "devices" collection. The purpose of this connector is only to abstract the calls to the MongoDB.Driver itself. In our endpoints, we still need to form the documents there; this connector will serve to accept those documents and perform those operations.

This story only concerns the "devices" collection. The "users" and "device-types" collections are not in this story's scope.

This PR also contains unit tests for the MongoConnector. All tests have passed on my local machine, and using the beima.yml GitHub Action.

I am aware this is a large pull request, but since all of this work is closely related, is an initial implementation, and builds on each other, it is difficult and highly time consuming to get these separated out into smaller PRs.

### DIFF
--- a/BEIMA.Backend.Test/MongoConnectorTest.cs
+++ b/BEIMA.Backend.Test/MongoConnectorTest.cs
@@ -37,32 +37,11 @@ namespace BEIMA.Backend.Test
                     Environment.SetEnvironmentVariable(setting.Key, setting.Value);
                 }
             }
-            catch (Exception ex)
+            catch (DirectoryNotFoundException ex)
             {
                 Console.WriteLine(ex.ToString());
             }
             
-        }
-
-        [OneTimeTearDown]
-        public void OneTimeTearDown()
-        {
-        }
-
-        [SetUp]
-        public void SetUp()
-        {
-        }
-
-        [TearDown]
-        public void TearDown()
-        {
-        }
-
-        [Test]
-        public void InitialState_Action_ExpectedResult()
-        {
-            Assert.IsTrue(true);
         }
 
         [Test]
@@ -74,10 +53,10 @@ namespace BEIMA.Backend.Test
             var connector3 = MongoConnector.Instance;
             var connector4 = MongoConnector.Instance;
             var connector5 = MongoConnector.Instance;
-            Assert.IsTrue(connector1 == connector2);
-            Assert.IsTrue(connector1 == connector3);
-            Assert.IsTrue(connector1 == connector4);
-            Assert.IsTrue(connector1 == connector5);
+            Assert.That(connector1, Is.EqualTo(connector2));
+            Assert.That(connector1, Is.EqualTo(connector3));
+            Assert.That(connector1, Is.EqualTo(connector4));
+            Assert.That(connector1, Is.EqualTo(connector5));
         }
 
         [Test]
--- a/BEIMA.Backend.Test/MongoConnectorTest.cs
+++ b/BEIMA.Backend.Test/MongoConnectorTest.cs
@@ -81,37 +81,17 @@ namespace BEIMA.Backend.Test
         }
 
         [Test]
-        public void ConnectorCreated_GetDeviceGivenInvalidDeviceId_NullReturned()
+        public void ConnectorNotCreated_CallConstructorOnce_InstanceConnectedToCorrectEnvironment()
         {
             var mongo = MongoConnector.Instance;
-            //This is a valid ObjectId, but this is not in the database
-            var doc = mongo.GetDevice(new ObjectId("61f4882f42dab933544849d3"));
-            Assert.IsNull(doc);
-        }
-
-        [Test]
-        public void ConnectorCreated_DeleteInvalidObject_FalseReturned()
-        {
-            var mongo = MongoConnector.Instance;
-            //This is a valid ObjectId, but this is not in the database
-            var result = mongo.DeleteDevice(new ObjectId("61f4882f42dab933544849d3"));
-            Assert.IsFalse(result);
-        }
-
-        [Test]
-        public void ConnectorCreated_InsertNull_NullReturned()
-        {
-            var mongo = MongoConnector.Instance;
-            var result = mongo.InsertDevice(null);
-            Assert.IsNull(result);
-        }
-
-        [Test]
-        public void ConnectorCreated_UpdateNull_NullReturned()
-        {
-            var mongo = MongoConnector.Instance;
-            var result = mongo.UpdateDevice(null);
-            Assert.IsFalse(result);
+            if (Environment.GetEnvironmentVariable("CurrentEnv") == "dev-local")
+            {
+                Assert.IsTrue(mongo.CurrentServerType == ServerType.Local);
+            }
+            else
+            {
+                Assert.IsTrue(mongo.CurrentServerType == ServerType.Cloud);
+            }
         }
 
         [Test]
@@ -150,20 +130,6 @@ namespace BEIMA.Backend.Test
             var retrievedDocs = mongo.GetAllDevices();
             Assert.IsNotNull(retrievedDocs);
             Assert.IsTrue(retrievedDocs is List<BsonDocument>);
-        }
-
-        [Test]
-        public void ConnectorNotCreated_CallConstructorOnce_InstanceConnectedToCorrectEnvironment()
-        {
-            var mongo = MongoConnector.Instance;
-            if (Environment.GetEnvironmentVariable("CurrentEnv") == "dev-local")
-            {
-                Assert.IsTrue(mongo.CurrentServerType == ServerType.Local);
-            }
-            else
-            {
-                Assert.IsTrue(mongo.CurrentServerType == ServerType.Cloud);
-            }
         }
 
         [Test]
@@ -221,6 +187,40 @@ namespace BEIMA.Backend.Test
             doc.AddRange(new BsonDocument { { "updatedDeviceField", "123" } });
             var updateResult = mongo.UpdateDevice(doc);
             Assert.IsTrue(updateResult);
+        }
+
+        [Test]
+        public void ConnectorCreated_GetDeviceGivenInvalidDeviceId_NullReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            //This is a valid ObjectId, but this is not in the database
+            var doc = mongo.GetDevice(new ObjectId("61f4882f42dab933544849d3"));
+            Assert.IsNull(doc);
+        }
+
+        [Test]
+        public void ConnectorCreated_DeleteInvalidObject_FalseReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            //This is a valid ObjectId, but this is not in the database
+            var result = mongo.DeleteDevice(new ObjectId("61f4882f42dab933544849d3"));
+            Assert.IsFalse(result);
+        }
+
+        [Test]
+        public void ConnectorCreated_InsertNull_NullReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            var result = mongo.InsertDevice(null);
+            Assert.IsNull(result);
+        }
+
+        [Test]
+        public void ConnectorCreated_UpdateNull_NullReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            var result = mongo.UpdateDevice(null);
+            Assert.IsFalse(result);
         }
 
     }
--- a/BEIMA.Backend.Test/MongoConnectorTest.cs
+++ b/BEIMA.Backend.Test/MongoConnectorTest.cs
@@ -84,12 +84,12 @@ namespace BEIMA.Backend.Test
             };
             var insertResult = mongo.InsertDevice(doc);
             Assert.IsNotNull(insertResult);
-            Assert.IsTrue(insertResult is ObjectId);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
 
             //Test (retrieve the document)
             var retrievedDoc = mongo.GetDevice((ObjectId) doc["_id"]);
             Assert.IsNotNull(retrievedDoc);
-            Assert.IsTrue(retrievedDoc is BsonDocument);
+            Assert.That(retrievedDoc, Is.TypeOf(typeof(BsonDocument)));
         }
 
         [Test]
@@ -103,12 +103,12 @@ namespace BEIMA.Backend.Test
             };
             var insertResult = mongo.InsertDevice(doc);
             Assert.IsNotNull(insertResult);
-            Assert.IsTrue(insertResult is ObjectId);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
 
             //Test (retrieve all documents)
             var retrievedDocs = mongo.GetAllDevices();
             Assert.IsNotNull(retrievedDocs);
-            Assert.IsTrue(retrievedDocs is List<BsonDocument>);
+            Assert.That(retrievedDocs, Is.TypeOf(typeof(List<BsonDocument>)));
         }
 
         [Test]
@@ -122,7 +122,7 @@ namespace BEIMA.Backend.Test
             var result = mongo.InsertDevice(doc);
             Console.WriteLine(result.ToString());
             Assert.IsNotNull(result);
-            Assert.IsTrue(result is ObjectId);
+            Assert.That(result, Is.TypeOf(typeof(ObjectId)));
         }
 
         [Test]
@@ -137,7 +137,7 @@ namespace BEIMA.Backend.Test
             //Insert device
             var insertResult = mongo.InsertDevice(doc);
             Assert.IsNotNull(insertResult);
-            Assert.IsTrue(insertResult is ObjectId);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
 
             //Delete device
             bool deleteResult = false;
@@ -160,7 +160,7 @@ namespace BEIMA.Backend.Test
             //Insert device
             var insertResult = mongo.InsertDevice(doc);
             Assert.IsNotNull(insertResult);
-            Assert.IsTrue(insertResult is ObjectId);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
 
             //Update device
             doc.AddRange(new BsonDocument { { "updatedDeviceField", "123" } });
--- a/BEIMA.Backend.Test/MongoConnectorTest.cs
+++ b/BEIMA.Backend.Test/MongoConnectorTest.cs
@@ -98,20 +98,6 @@ namespace BEIMA.Backend.Test
         }
 
         [Test]
-        public void ConnectorNotCreated_CallConstructorOnce_InstanceConnectedToCorrectEnvironment()
-        {
-            var mongo = MongoConnector.Instance;
-            if (Environment.GetEnvironmentVariable("CurrentEnv") == "dev-local")
-            {
-                Assert.IsTrue(mongo.CurrentServerType == ServerType.Local);
-            }
-            else
-            {
-                Assert.IsTrue(mongo.CurrentServerType == ServerType.Cloud);
-            }
-        }
-
-        [Test]
         public void ConnectorCreated_GetDeviceGivenValidDeviceId_DeviceDocumentReturned()
         {
             var mongo = MongoConnector.Instance;
@@ -240,5 +226,12 @@ namespace BEIMA.Backend.Test
             Assert.IsNull(result);
         }
 
+        [Test]
+        public void ConnectorCreated_CheckIsConnected_TrueReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            var result = mongo.IsConnected();
+            Assert.IsTrue(result);
+        }
     }
 }
--- a/BEIMA.Backend.Test/MongoConnectorTest.cs
+++ b/BEIMA.Backend.Test/MongoConnectorTest.cs
@@ -68,6 +68,7 @@ namespace BEIMA.Backend.Test
         [Test]
         public void ConnectorNotCreated_CallConstructorFiveTimes_FiveInstancesAreEqual()
         {
+            //Tests to make sure singleton is working
             var connector1 = MongoConnector.Instance;
             var connector2 = MongoConnector.Instance;
             var connector3 = MongoConnector.Instance;
@@ -117,7 +118,7 @@ namespace BEIMA.Backend.Test
         public void ConnectorCreated_GetDeviceGivenValidDeviceId_DeviceDocumentReturned()
         {
             var mongo = MongoConnector.Instance;
-            //Test setup
+            //Setup (inserting a document)
             BsonDocument doc = new BsonDocument {
                 { "deviceTypeId", "a" },
                 { "serialNumber", "a12345"}
@@ -126,7 +127,7 @@ namespace BEIMA.Backend.Test
             Assert.IsNotNull(insertResult);
             Assert.IsTrue(insertResult is ObjectId);
 
-            //Test
+            //Test (retrieve the document)
             var retrievedDoc = mongo.GetDevice((ObjectId) doc["_id"]);
             Assert.IsNotNull(retrievedDoc);
             Assert.IsTrue(retrievedDoc is BsonDocument);
--- a/BEIMA.Backend.Test/MongoConnectorTest.cs
+++ b/BEIMA.Backend.Test/MongoConnectorTest.cs
@@ -21,16 +21,24 @@ namespace BEIMA.Backend.Test
         [OneTimeSetUp]
         public void OneTimeSetUp()
         {
-            string[] paths = { "..", "..", "..", "..", "BEIMA.Backend", "local.settings.json" };
-            string combinedPath = Path.Combine(paths);
-            string fullPath = Path.GetFullPath(combinedPath);
-            var settings = JsonConvert.DeserializeObject<LocalSettings>(
-                File.ReadAllText(fullPath));
-
-            foreach (var setting in settings.Values)
+            try
             {
-                Environment.SetEnvironmentVariable(setting.Key, setting.Value);
+                string[] paths = { "..", "..", "..", "..", "BEIMA.Backend", "local.settings.json" };
+                string combinedPath = Path.Combine(paths);
+                string fullPath = Path.GetFullPath(combinedPath);
+                var settings = JsonConvert.DeserializeObject<LocalSettings>(
+                    File.ReadAllText(fullPath));
+
+                foreach (var setting in settings.Values)
+                {
+                    Environment.SetEnvironmentVariable(setting.Key, setting.Value);
+                }
             }
+            catch (Exception ex)
+            {
+                Console.WriteLine(ex.ToString());
+            }
+            
         }
 
         [OneTimeTearDown]
--- a/BEIMA.Backend.Test/MongoConnectorTest.cs
+++ b/BEIMA.Backend.Test/MongoConnectorTest.cs
@@ -1,0 +1,109 @@
+﻿using NUnit.Framework;
+using BEIMA.Backend.MongoService;
+using System.IO;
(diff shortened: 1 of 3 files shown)


BEIMA.Backend.Test/MongoConnectorTest.cs
```diff
@@ -0,0 +1,227 @@
...
+                string fullPath = Path.GetFullPath(combinedPath);
+                var settings = JsonConvert.DeserializeObject<LocalSettings>(
+                    File.ReadAllText(fullPath));
+
+                foreach (var setting in settings.Values)
+                {
+                    Environment.SetEnvironmentVariable(setting.Key, setting.Value);
+                }
+            }
+            catch (Exception ex)
```
josephmoore99, line 40: We should catch the specific exception type rather than any exception, just to be safe

BEIMA.Backend.Test/MongoConnectorTest.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+        }
+
+        [Test]
+        public void InitialState_Action_ExpectedResult()
```
josephmoore99, line 63: I think we can remove this, unless we want to leave it as an example

BEIMA.Backend.Test/MongoConnectorTest.cs
```diff
@@ -0,0 +1,227 @@
...
+            //Tests to make sure singleton is working
+            var connector1 = MongoConnector.Instance;
+            var connector2 = MongoConnector.Instance;
+            var connector3 = MongoConnector.Instance;
+            var connector4 = MongoConnector.Instance;
+            var connector5 = MongoConnector.Instance;
+            Assert.IsTrue(connector1 == connector2);
+            Assert.IsTrue(connector1 == connector3);
+            Assert.IsTrue(connector1 == connector4);
+            Assert.IsTrue(connector1 == connector5);
```
josephmoore99, line 80:
```suggestion
            Assert.That(connector1, Is.EqualTo(connector2));
            Assert.That(connector1, Is.EqualTo(connector3));
            Assert.That(connector1, Is.EqualTo(connector4));
            Assert.That(connector1, Is.EqualTo(connector5));
```
Not a big issue, but usually it's better to use this format because it will log a much less vague error message if it fails.

BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -0,0 +1,192 @@
...
+            }
+            else
+            {
+                CurrentServerType = ServerType.Cloud;
+                credentials = Environment.GetEnvironmentVariable("AzureCosmosConnection");
+            }
+            client = new MongoClient(credentials);
+        }
+
+        /*
```
josephmoore99, line 45: I have been writing my method/class headers using a different format, we should likely try to come up with a standard format. I think the one that I use is pretty nice because you only need to type '///' in VS and it will automatically put in a template for you.

BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -0,0 +1,192 @@
...
+            {
+                throw new Exception("MongoConnector is not currently connected");
+            }
+
+            try
+            {
+                var db = client.GetDatabase(dbName);
+                var devices = db.GetCollection<BsonDocument>(deviceCollection);
+                devices.InsertOne(doc);
+                return (ObjectId) doc["_id"];
```
josephmoore99, line 72: Will this match the id in the database?

BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        /*
+         * Inserts a device into the "devices" collection
+         * Parameter: BsonDocument that contains the fully formed device document (including all required and optional fields)
+         * Returns: ObjectId of the newly inserted object if successful, null if failed
+         */
+        public ObjectId? InsertDevice(BsonDocument doc)
+        {
+            if (!IsConnected())
```
josephmoore99, line 62: Since this check is so common, we could probably just put it in a method.

BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -0,0 +1,192 @@
...
+        //Public class members
+        public ServerType CurrentServerType { get; }
+
+        //Contains instance variables
+        private readonly MongoClient client = null;
+        private static readonly Lazy<MongoConnector> instance = new(() => new MongoConnector());
+
+        //Environment variables
+        private readonly string dbName = Environment.GetEnvironmentVariable("DatabaseName");
+        private readonly string deviceCollection = Environment.GetEnvironmentVariable("DeviceCollectionName");
```
josephmoore99, line 22: I think this would be better to put in a constants class rather than an environment variable?

BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -0,0 +1,192 @@
...
+                throw new Exception("MongoConnector is not currently connected");
+            }
+
+            var filter = Builders<BsonDocument>.Filter.Empty;
+
+            try
+            {
+                var db = client.GetDatabase(dbName);
+                var devices = db.GetCollection<BsonDocument>(deviceCollection);
+                var docs = devices.Find(filter).ToList();
```
josephmoore99, line 126: Is this filter necessary?

BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -0,0 +1,192 @@
...
+            }
+
+            var filter = Builders<BsonDocument>.Filter.Eq("_id", objectId);
+
+            try
+            {
+                var db = client.GetDatabase(dbName);
+                var devices = db.GetCollection<BsonDocument>(deviceCollection);
+                var result = devices.DeleteOne(filter);
+                return result.DeletedCount > 0;
```
josephmoore99, line 155: I think it would be better to return the list of deleted devices instead?

BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -0,0 +1,192 @@
...
+            }
+
+            try
+            {
+                ObjectId objectId = (ObjectId)doc["_id"];
+                var filter = Builders<BsonDocument>.Filter.Eq("_id", objectId);
+                var db = client.GetDatabase(dbName);
+                var devices = db.GetCollection<BsonDocument>(deviceCollection);
+                var result = devices.ReplaceOne(filter, doc);
+                return result.ModifiedCount > 0;
```
josephmoore99, line 183: Likewise, we could return the updated device here instead?

BEIMA.Backend.Test/MongoConnectorTest.cs
josephmoore99: Can do this for the other ones as well. Basically, instead of saying "the assertion was false", it will say something like "connector1 was not equal to connector2"

BEIMA.Backend/MongoService/MongoConnector.cs
AshlynAdamson: I think if the argument passed into the InsertOne method is pass by reference, then it should update and match

BEIMA.Backend/MongoService/MongoConnector.cs
AshlynAdamson: We could probably even have a method to write the exception message and return null as well since that is used quite a bit too

BEIMA.Backend.Test/MongoConnectorTest.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            var mongo = MongoConnector.Instance;
+            //Setup (inserting a document)
+            BsonDocument doc = new BsonDocument {
+                { "deviceTypeId", "a" },
+                { "serialNumber", "a12345"}
+            };
+            var insertResult = mongo.InsertDevice(doc);
+            Assert.IsNotNull(insertResult);
+            Assert.IsTrue(insertResult is ObjectId);
```
josephmoore99, line 108:
```suggestion
            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
```
Again, just me being picky, but I generally like to avoid using Assert.IsTrue if possible.

BEIMA.Backend.Test/MongoConnectorTest.cs
```diff
@@ -0,0 +1,227 @@
...
+
+            //Test (retrieve all documents)
+            var retrievedDocs = mongo.GetAllDevices();
+            Assert.IsNotNull(retrievedDocs);
+            Assert.IsTrue(retrievedDocs is List<BsonDocument>);
```
josephmoore99, line 132: We should check individual list items as well.

BEIMA.Backend/MongoService/MongoConnector.cs
AshlynAdamson: I like this idea too, makes the return types more consistent

josephmoore99 (changes requested):
Great stuff! Cool to see that we have some basic DB functionality now!

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: I think I will leave this in, it is easy to see the format and this test is a guaranteed pass. If it fails, then we know something is wrong with the framework.

BEIMA.Backend.Test/MongoConnectorTest.cs
AshlynAdamson: If we can get a more detailed log then I think that is a good change to make

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: When InsertOne() is called, upon insertion, the passed in "doc" will automatically be updated to contain the new ObjectId.

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: The List<BsonDocument> is already type safe for BsonDocuments.

BEIMA.Backend/MongoService/MongoConnector.cs
Kenny-Miller: I don't think its really necessary. The ui would already have the updated state and I don't think the rest of the api would need to use that object for anything.

BEIMA.Backend.Test/MongoConnectorTest.cs
josephmoore99: I meant like, checking to see that the items in the list have the actual data we expect them to have.

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: I don't think it is necessary either, since you would already have a copy of the document itself in memory in order to call this method in the first place. This way, by returning a boolean, we can keep sanity checking very easy and only have to do "if(result)". In essence, it would be redundant to return a copy of the document again.

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: To delete a device, you would need to already have a copy of the object itself to call this method. Returning the deleted device is redundant since we already have a reference to the device. Having it return a boolean makes sanity checking much easier since we can do something like "if(result)".

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: Yes, an empty filter will return all objects. This is the generally accepted way to do this.

BEIMA.Backend.Test/MongoConnectorTest.cs
Kenny-Miller: I like how this reads better too.

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: I see it as an environment variable. The user does not have to use "beima" or "devices" as the collection names, it is completely dependent on how they want their environment set up. This is an open source project, so people might use their own name in place of "beima", or they might want to use their own language in place of "devices".

BEIMA.Backend/MongoService/MongoConnector.cs
josephmoore99: I think it's good for if we need to verify anything at the very least. I don't see too much harm in it.

BEIMA.Backend.Test/MongoConnectorTest.cs
josephmoore99: Yeah, usually Assert.That(object, Is....) is my go to format for this.

BEIMA.Backend.Test/MongoConnectorTest.cs
AshlynAdamson: I think we can remove it. If this is a valuable example, we should document this in a Testing Guide/Coding Standards that we can create and use as reference. If something is wrong with the framework, I think the `dotnet test` command will catch it before it runs any tests. Or we make a separate test the verifies the framework is working. It adds unneeded coupling to this test file.

If we agree to keep it, we should add a comment stating why. If a new developer were to come in and see this test case, it wouldn't be very clear as to why we left it. Also, we should remove the OneTimeTearDown, Setup, and TearDown methods since they are doing nothing. Unless they are needed to run the test in the framework, then we should add a comment indicating that.

BEIMA.Backend/MongoService/MongoConnector.cs
AshlynAdamson: I think this is a good point to bring up in retro. Being consistent throughout our code base is important to the maintainability of BEIMA

BEIMA.Backend/MongoService/MongoConnector.cs
josephmoore99: The beima one makes sense to me, but I wouldn't think we would want to change the device name? I guess I'm just unfamiliar with this type of database.

roguedeleter: Collections in MongoDB are equivalent to tables in SQL. If someone speaking a language other than English wanted to change the name of the "devices" collection to whatever "devices" is in their language, they should have to ability to do so. Mongo is very flexible when it comes to these type of things.

BEIMA.Backend/MongoService/MongoConnector.cs
AshlynAdamson: I agree, doesn't add any harm. We can use `if(result != NULL)`, that doesn't add any additional code. Thinking ahead to what BEIMA could be, what if we want undo functionality if a user deletes an entry they didn't mean to? What about keep a session log for a user and having to track all the entries a user deleted? We can do more with an object than we can with a boolean. Just some ideas

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: @josephmoore99 @AshlynAdamson should this be changed now, or should it wait until retro?

BEIMA.Backend.Test/MongoConnectorTest.cs
Kenny-Miller: I think that we should remove it as its just unnecessary code. I agree with Ashlyn that we could put it in a document if we want to keep it around.

BEIMA.Backend/MongoService/MongoConnector.cs
josephmoore99: Yeah, I think I was thinking about this being related to using remove() on a list, but given that we actually need to have the device info to use this, returning true/false seems reasonable.

BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -0,0 +1,192 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using MongoDB.Driver;
+using MongoDB.Bson;
+
+namespace BEIMA.Backend.MongoService
+{
+    public sealed class MongoConnector
```
josephmoore99, line 11: I think it's important to have class headers as well, for maintainability.

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: Fixed

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: Fixed [6f34fb6](https://github.com/cs481-ekh/s22-beima/pull/67/commits/6f34fb69b240040dce07b88ac263f3df03d88033)

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: Fixed [6f34fb6](https://github.com/cs481-ekh/s22-beima/pull/67/commits/6f34fb69b240040dce07b88ac263f3df03d88033)

BEIMA.Backend.Test/MongoConnectorTest.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            var mongo = MongoConnector.Instance;
+            if (Environment.GetEnvironmentVariable("CurrentEnv") == "dev-local")
+            {
+                Assert.IsTrue(mongo.CurrentServerType == ServerType.Local);
+            }
+            else
+            {
+                Assert.IsTrue(mongo.CurrentServerType == ServerType.Cloud);
+            }
```
AshlynAdamson, line 73: We should make this test only expect one type, that way if it fails we know exactly what environment it was set in. Since the MongoConnector is a singleton, I think we can create another test class that tests when this environment variable is not "dev-local"

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: Fixed [f86d5e8](https://github.com/cs481-ekh/s22-beima/pull/67/commits/f86d5e8a2e85d6c7f43601eb50957275b48efcd8)

BEIMA.Backend.Test/MongoConnectorTest.cs
AshlynAdamson: I agree, the more testing we can do the better.

BEIMA.Backend/MongoService/MongoConnector.cs
AshlynAdamson: If you agree with the header change, then we can do it now. But if you like what you had, then I think leaving it until retro is fine, gives us a chance to discuss about it

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: Fixed [d10ddbb](https://github.com/cs481-ekh/s22-beima/pull/67/commits/d10ddbbb704bdd673bd1c99f2a7b83230349ca7b)

BEIMA.Backend.Test/MongoConnectorTest.cs
AshlynAdamson: We could even have another test class that doesn't initialize the client to verify the correct error message is given

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: Fixed b8e46ae

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: Fixed c07b290

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: I will keep it the same, it will return a boolean. From what I found, delete operations typically don't return anything. This way, at least we can see if it failed or not by seeing if it returned true/false.

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: Fixed d12ac92

BEIMA.Backend.Test/MongoConnectorTest.cs
```diff
@@ -0,0 +1,206 @@
...
+            var updateResult = mongo.UpdateDevice(doc);
+            Assert.IsTrue(updateResult);
+        }
+
+        [Test]
+        public void ConnectorCreated_GetDeviceGivenInvalidDeviceId_NullReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            //This is a valid ObjectId, but this is not in the database
+            var doc = mongo.GetDevice(new ObjectId("61f4882f42dab933544849d3"));
```
AshlynAdamson, line 176: Will there be a situation where this could become an actual ID in the database? Are we using a specific test database?

BEIMA.Backend.Test/MongoConnectorTest.cs
```diff
@@ -0,0 +1,206 @@
...
+                { "serialNumber", "a12345"}
+            };
+            var insertResult = mongo.InsertDevice(doc);
+            Assert.IsNotNull(insertResult);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
+
+            //Test (retrieve all documents)
+            var retrievedDocs = mongo.GetAllDevices();
+            Assert.IsNotNull(retrievedDocs);
+            Assert.That(retrievedDocs, Is.TypeOf(typeof(List<BsonDocument>)));
```
AshlynAdamson, line 111: Could there be a situation where there are more than two entries in the database and this test case takes longer to run dur to that?

BEIMA.Backend.Test/MongoConnectorTest.cs
```diff
@@ -0,0 +1,206 @@
...
+            var result = mongo.DeleteDevice(new ObjectId("61f4882f42dab933544849d3"));
+            Assert.IsFalse(result);
+        }
+
+        [Test]
+        public void ConnectorCreated_InsertNull_NullReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            var result = mongo.InsertDevice(null);
+            Assert.IsNull(result);
```
AshlynAdamson, line 194: Do we want NULL to be returned if it is inserted or should we throw an exception?

BEIMA.Backend/MongoService/MongoConnector.cs
josephmoore99: I think the '///' is the standard for C#. Plus it also helps provide method info when hovering over a method call or class reference in VS.

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: I wanted null to be returned so we don't have to wrap everything in our endpoints to be wrapped in try/catch blocks. Instead, the exception is abstracted out and we have something easier to check.

BEIMA.Backend.Test/MongoConnectorTest.cs
josephmoore99: I think that the chances of this actually happening with a guid are so miniscule, it shouldn't be an issue. 1 in 2^128 I believe.

roguedeleter: There shouldn't be, unless we are extremely unlucky. The probability of this occurring is 16^24 I think? I had to use some hexadecimal value of length 24, otherwise ObjectId would refuse to instantiate.

BEIMA.Backend.Test/MongoConnectorTest.cs
josephmoore99: I think discussing how we want to do exception handling would be a good thing to bring up in the retro as well.

BEIMA.Backend/MongoService/MongoConnector.cs
josephmoore99: Ok, as long as it's the standard I'm cool with it.

BEIMA.Backend.Test/MongoConnectorTest.cs
josephmoore99: I think we would ideally be clearing the database after each test, that way tests don't interfere with one another. But if we aren't hosting the database on the local build machine, we might want to consider doing that or setting up a mock database class. I think either task would have to be done next sprint.

BEIMA.Backend.Test/MongoConnectorTest.cs
```diff
@@ -0,0 +1,206 @@
...
+namespace BEIMA.Backend.Test
+{
+    [TestFixture]
+    public class MongoConnectorTest
+    {
+        //private MongoConnector? mongo;
+        class LocalSettings
+        {
+            public bool IsEncrypted { get; set; }
+            public Dictionary<string, string> Values { get; set; }
```
josephmoore99, line 18: This being null is making the github thing upset.

roguedeleter: Fixed be69b94

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: Fixed be69b94

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: Having separate collections solves this issue entirely, be69b94

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: This test has been removed after an offline conversation. This should be an integration or functional test, it does not make sense for it to be a unit test. afb7770

josephmoore99:
I think what we have here is good for now. I think we might want to discuss how we will want to test functionality involving database calls in unit tests during the retro. Since this class is the class that is the closest layer to the database, I think it is fine to have it connecting to the actual database. However, I think that any of our other unit tests should be using a mock database client, because I can see having to reset the database for every test being a costly process. We can likely use an interface that both our mock db client and real db client can inherit from, so that they use the same methods. Likely a task for next sprint though.

BEIMA.Backend.Test/MongoConnectorTest.cs
roguedeleter: After a discussion with @AshlynAdamson offline, we have determined that this is better as a functional test. The collection is in an uncertain state when this unit test is run. Because of this uncertainty, it is better if we make it an FT so the state is actually certain.

josephmoore99 (approved):

AshlynAdamson (approved):

Kenny-Miller (approved):

PR #106, opened 2022-02-19
Creating NavBar and PageTitle shared components

### LINKED ISSUE #83: Create sidebar navigation shared component
References user story #78 

As a user, I need a sidebar that I can use to navigate to the different pages.

### LINKED ISSUE #84: Create a toolbar component that displays the current page
References user story #78

As a user, I need a toolbar that displays what 'page' I am on.

AshlynAdamson:
Closes #83 
Closes #84 

Adding a NavBar and PageTitle. I tried to follow our mockup as best we could. There are still some sizing/scaling issues, but I wanted to get it up for review and get some feedback. Also added to the routing tests as well. There was also something funky when it would load the help page, maybe something to do with how the we pass the page to the Outlet tag? @Kenny-Miller, do you have an idea of what is causing it?

Here's a video showing how it behaves so far

https://user-images.githubusercontent.com/58532471/154816719-63029b21-23ef-4d59-83b4-d0d4bab89126.mp4


BEIMA.Client/cypress/integration/Router.spec.js
```diff
@@ -4,20 +4,54 @@ describe("Router Doesn't Redirect On Valid Pages", () =>{
   it('Visits Root', () =>{
     cy.visit('http://localhost:3000')
     cy.url().should('eq','http://localhost:3000/')
+    cy.contains('Devices')
```
josephmoore99, line 7: cy.contains by itself seems like it might be an expensive operation based on the little bit I looked up on it. Should we be identifying the DOM object with cy.get().contains() rather than just using the contains method?

BEIMA.Client/src/shared/NavBar.js
```diff
@@ -0,0 +1,35 @@
+import logo from './BSU-logo.png';
+import { Row, Col} from 'react-bootstrap';
+import './shared.css';
+
+const NavBar = () => {
```
josephmoore99, line 5: We might consider adding ids on components so that we can uniquely identify them in testing.

AshlynAdamson:
The reason the help page looks weird is because the height doesn't fill the whole page, so that will be fixed with styling

BEIMA.Client/src/shared/NavBar.js
AshlynAdamson: There will only be one nav bar and one page title throughout, I think in this case we don't need IDs on the nav bar and page title

BEIMA.Client/cypress/integration/Router.spec.js
AshlynAdamson: Yeah that's a good idea, that way we can be more specific in our tests too

thessbsu:
Does the logo hugging the very top look weird to anyone else? I think it might look better with the top of the 'B' at the same height as the top border around 'Home' in the nav bar

AshlynAdamson:
> Does the logo hugging the very top look weird to anyone else? I think it might look better with the top of the 'B' at the same height as the top border around 'Home' in the nav bar

Yeah good point, didn't really notice it before. I added a top margin and here's what it looks like now 
image

josephmoore99 (approved):

AshlynAdamson:
- [ ] I noticed the background color of the nav bar doesn't reach the whole page, gonna figure that out

BEIMA.Client/src/shared/NavBar.js
```diff
@@ -0,0 +1,35 @@
...
+    return (
+        <div className="sharedNavBar">
+            <Col>
```
roguedeleter, line 8: I would wrap these in a "Container" element, according to this documentation: https://react-bootstrap.github.io/layout/grid/

BEIMA.Client/src/shared/NavBar.js
```diff
@@ -0,0 +1,35 @@
...
+                <Row>
+                    <Col className="pageLinks">
+                        <Row className="pageLink">
+                            <a href="Devices" className="sharedText">Devices</a>
+                        </Row>
+                        <Row className="pageLink">
+                            <a href="Device Types" className="sharedText">Device Types</a>
+                        </Row> 
+                        <Row className="pageLink">
+                            <a href="Add Device" className="sharedText">Add Device {"\xa0\xa0"}+</a>
```
roguedeleter, line 20: Instead of putting "+" at the end, let's try and put it at the beginning. I envision that all of these sidebar links will have their own unique icon, so it will look a lot more normal in the future. This also avoids having to use nonbreaking spaces.

roguedeleter (changes requested):
Good job, looks great!

roguedeleter (approved):

josephmoore99 (approved):

PR #107, opened 2022-02-19
More workflow changes

AshlynAdamson:

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-#Update the apt package manager
-sudo apt update || { echo "Updating apt failed"; exit 1; }
 # Install software-properties-common apt-transport-https for wget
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-#Update the apt package manager
-sudo apt update || { echo "Updating apt failed"; exit 1; }
```
josephmoore99, line 4: Do we think we might ever need this in the future?

josephmoore99 (approved):

build.sh
AshlynAdamson: Hmm, hard to say. If the hosting VM ever needs this command to be ran, then someone can manually do it

thessbsu (approved):

PR #108, opened 2022-02-19
Add device get functionality and tests

### LINKED ISSUE #98: Add device GET endpoint
References user story #77

As a user, I need to be able to view a device entry so that I can see it's details.

josephmoore99:
Closes #98

Adds functionality to retrieve a device from the database using an id from a query string

### DIFF
--- a/BEIMA.Backend.Test/DeviceTest.cs
+++ b/BEIMA.Backend.Test/DeviceTest.cs
@@ -23,7 +23,7 @@ namespace BEIMA.Backend.Test
 
             // Setup mock database client.
             Mock<IMongoConnector> mockDb = new Mock<IMongoConnector>();
-            var testGuid = "1234567890abcdef1234567890abcdef";
+            var testGuid = "1234567890abcdef12345678";
             mockDb.Setup(mock => mock.GetDevice(It.Is<ObjectId>(oid => oid == new ObjectId(testGuid))))
                   .Returns<BsonDocument>(null)
                   .Verifiable();
@@ -46,7 +46,8 @@ namespace BEIMA.Backend.Test
         }
 
         [TestCase("")]
-        [TestCase("1234")]
+        [TestCase("xxx")]
+        [TestCase("123")]
         public async Task InvalidId_GetDevice_ReturnsInvalidId(string id)
         {
             // ARRANGE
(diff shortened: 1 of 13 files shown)


BEIMA.Backend/MongoService/MongoDefinition.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// <remarks>
+    /// The purpose of this class is to allow us to switch out the default
+    /// Mongo connector with a mock Mongo connector for testing.
+    /// </remarks>
+    public static class MongoDefinition
+    {
+        /// <summary>
+        /// Gets and sets the currently defined instance of the Mongo connector
+        /// </summary>
+        public static IMongoConnector MongoInstance { get; set; } = MongoConnector.Instance;
```
josephmoore99, line 15: Thought this was the best way to do this, but I wanted to be able to switch out the running instance of the MongoConnector with a Mock instance. So the endpoint will call this instead of MongoConnector.Instance, and that way it will use whichever instance is set here. By default it is set to the actual mongo instance.

BEIMA.Backend.Test/DeviceTest.cs
```diff
@@ -0,0 +1,112 @@
...
+        public async Task IdNotInDatabase_GetDevice_ReturnsNotFound()
+        {
+            // ARRANGE
+
+            // Setup mock database client.
+            Mock<IMongoConnector> mockDb = new Mock<IMongoConnector>();
+            var testGuid = "1234567890abcdef1234567890abcdef";
+            mockDb.Setup(mock => mock.GetDevice(It.Is<ObjectId>(oid => oid == new ObjectId(testGuid))))
+                  .Returns<BsonDocument>(null)
+                  .Verifiable();
```
josephmoore99, line 29: This can be used as a sort of reference for how to do a mock client. You can set individual methods to return specific values depending on the input parameters.

BEIMA.Backend/Device.cs
```diff
@@ -0,0 +1,76 @@
...
+                    response.StatusCode = (int)HttpStatusCode.NotFound;
+                    return response;
+                }
+
+                // Return the device.
+                response = new OkObjectResult(doc);
+
+            }
+            // Process as a device POST request for creating, updating, or deleting a device.
+            else if (string.Equals(req.Method, "post", StringComparison.OrdinalIgnoreCase))
```
josephmoore99, line 62: I am going to wait on doing the post stuff for device in another task. Also will need the factories set up for this to work I imagine.

AshlynAdamson (approved):
I was able to hit this endpoint using Postman, looks good! I like the UnitTestBase file, I'm sure that will get used frequently

thessbsu (approved):

AshlynAdamson (approved):

Issue #80, opened 2022-02-17
Create POCOs to Properly Format BsonDocuments


roguedeleter:
#2 

This story involves implementing a Factory class to properly format BsonDocuments in the correct schema according to the schemas located in BEIMA.DB.Schemas.

Essentially, the Factory will take in raw parameters (string, int, etc), and shove it all into a BsonDocument. It will return the correctly formatted BsonDocument, and that object will be passed in to the MongoConnector for CRUD operations.

The endpoint will be what is calling the factory. Here is the data flow:

1. The endpoint receives a POST request.
2. The endpoint will parse out the POST request body, and extract fields such as serial number, model number, etc.
3. The endpoint will call the Factory, take all the parameters from step 2, and pass them all as parameters into the Factory.
4. The Factory will take all these parameters and structure them properly into a BsonDocument. This **MUST** conform to the schemas located in BEIMA.DB.Schemas.
5. The Factory will return the BsonDocument to the endpoint.
6. The endpoint will call the MongoConnector, and pass in the properly structured BsonDocument from step 5.
7. The MongoConnector will perform the raw CRUD operation, and insert/update the document in Mongo.

UPDATE: Instead of a Factory, I have chosen to use POCOs (Plain Old C# Objects) instead. A true factory class is a bit overkill for this.

- [x] #110
- [x] #111
- [x] #112
- [x] #113
- [x] #229

PR #114, opened 2022-02-22
Create C# object for BaseDevice

### LINKED ISSUE #110: Create C# object for BaseDevice
#80 

We need a C# object for a BaseDevice. This should allow us to convert data into a BsonDocument.

roguedeleter:
Closes #110 

This PR implements a BaseDevice object that we can use to convert raw Device data into a BsonDocument. This is part of the "Factory" story #80, but I decided against implementing a Factory class. This type of work is easily done using a C# object instead, since there exist methods out there that automatically serialize objects into BSON.

### DIFF
--- a/BEIMA.Backend.Test/MongoService/BaseDeviceTest.cs
+++ b/BEIMA.Backend.Test/MongoService/BaseDeviceTest.cs
@@ -7,9 +7,36 @@ namespace BEIMA.Backend.MongoService.Test
     [TestFixture]
     public class BaseDeviceTest
     {
+        //Instance variables
+        readonly ObjectId validObjId = ObjectId.GenerateNewId();
+        readonly ObjectId validDeviceTypeId = ObjectId.GenerateNewId();
+        readonly string validDeviceTag = "DeviceTag123";
+        readonly string validManufacturer = "BSU";
+        readonly string validModelNum = "ABC-123";
+        readonly string validSerialNum = "ABCDEFG123456";
+        readonly int validYearManufactured = 2022;
+        readonly string validNotes = "Example of notes.";
+
+        readonly DateTime validDate = DateTime.Now.ToUniversalTime();
+        readonly string validUser = "User1";
+
+        readonly ObjectId validBuildingId = ObjectId.GenerateNewId();
+        readonly string validLocationNotes = "Example of location notes.";
+        readonly string validLatitude = "-000.21.21";
+        readonly string validLongitude = "-323.32.32";
+
+        readonly string validKey1 = "key1";
+        readonly string validKey2 = "key2";
+        readonly string validKey3 = "key3";
+        readonly string validValue1 = "value1";
+        readonly bool validValue2 = true;
+        readonly int validValue3 = 123;
+
+
         [OneTimeSetUp]
         public void OneTimeSetUp()
         {
+
         }
 
         [OneTimeTearDown]
@@ -30,21 +57,81 @@ namespace BEIMA.Backend.MongoService.Test
         [Test]
         public void InitialState_Action_ExpectedResult()
         {
-            var objId = new ObjectId("61f4882f42dab933544849d3");
-            var device = new BaseDevice(objId, "asd", "asd", "xcsd", "cvliokj", "asd", 1, "sadloifk");
-            device.AddField("asdfwefsdf", "test-value");
-            device.AddField("abc", "test2");
-            device.SetLastModified(DateTime.Now, "gooduser0");
-            device.SetLastModified(DateTime.Now, "gooduser1");
-            device.SetLastModified(DateTime.Now, "gooduser2");
-            Console.WriteLine(device.GetBsonDocument().ToString());
+            //var objId = new ObjectId("61f4882f42dab933544849d3");
+            //var device = new BaseDevice(objId, "asd", "asd", "xcsd", "cvliokj", "asd", 1, "sadloifk");
+            //device.AddField("asdfwefsdf", "test-value");
+            //device.AddField("abc", "test2");
+            //device.SetLastModified(DateTime.Now, "gooduser0");
+            //device.SetLastModified(DateTime.Now, "gooduser1");
+            //device.SetLastModified(DateTime.Now, "gooduser2");
+            //Console.WriteLine(device.GetBsonDocument().ToString());
 
 
-            var testDevice = new BaseDevice
+            //var testDevice = new BaseDevice
+            //{
+            //    DeviceTypeId = "dslkfj",
+            //    Fields = new BsonDocument()
+            //};
+        }
+
+        [Test]
+        public void BaseDeviceNotInstantiated_InstantiateUsingFullConstructor_BaseDeviceInstantiatedWithCorrectValues()
+        {
+            var device = new BaseDevice(validObjId, validDeviceTypeId, validDeviceTag, validManufacturer, validModelNum, validSerialNum, validYearManufactured, validNotes);
+            Assert.That(device.Id, Is.EqualTo(validObjId));
+            Assert.That(device.DeviceTypeId, Is.EqualTo(validDeviceTypeId));
+            Assert.That(device.DeviceTag, Is.EqualTo(validDeviceTag));
+            Assert.That(device.Manufacturer, Is.EqualTo(validManufacturer));
+            Assert.That(device.ModelNum, Is.EqualTo(validModelNum));
+            Assert.That(device.SerialNum, Is.EqualTo(validSerialNum));
+            Assert.That(device.YearManufactured, Is.EqualTo(validYearManufactured));
+            Assert.That(device.Notes, Is.EqualTo(validNotes));
+
+            device.SetLastModified(validDate, validUser);
+            var lastModified = device.LastModified;
+            Assert.That((DateTime)lastModified.GetElement("date").Value, Is.EqualTo(validDate).Within(5).Seconds);
+            Assert.That((string)lastModified.GetElement("user").Value, Is.EqualTo(validUser));
+
+            device.SetLocation(validBuildingId, validLocationNotes, validLatitude, validLongitude);
+            var location = device.Location;
+            Assert.That((ObjectId)location.GetElement("buildingId").Value, Is.EqualTo(validBuildingId));
+            Assert.That((string)location.GetElement("notes").Value, Is.EqualTo(validLocationNotes));
+            Assert.That((string)location.GetElement("latitude").Value, Is.EqualTo(validLatitude));
+            Assert.That((string)location.GetElement("longitude").Value, Is.EqualTo(validLongitude));
+
+            device.AddField(validKey1, validValue1);
+            device.AddField(validKey2, validValue2);
+            device.AddField(validKey3, validValue3);
+            var fields = device.Fields;
+            Assert.That((string)fields.GetElement(validKey1).Value, Is.EqualTo(validValue1));
+            Assert.That((bool)fields.GetElement(validKey2).Value, Is.EqualTo(validValue2));
+            Assert.That((int)fields.GetElement(validKey3).Value, Is.EqualTo(validValue3));
+        }
+
+        [Test]
+        public void BaseDeviceNotInstantiated_InstantiateUsingObjectInitializer_BaseDeviceInstantiatedWithCorrectValues()
+        {
+            var device = new BaseDevice
             {
-                DeviceTypeId = "dslkfj",
-                Fields = new BsonDocument()
+                Id = validObjId,
+                DeviceTypeId = validDeviceTypeId,
+                DeviceTag = validDeviceTag,
+                Manufacturer = validManufacturer,
+                ModelNum = validModelNum,
+                SerialNum = validSerialNum,
+                YearManufactured = validYearManufactured,
+                Notes = validNotes
             };
+            Assert.That(device.Id, Is.EqualTo(validObjId));
+            Assert.That(device.DeviceTypeId, Is.EqualTo(validDeviceTypeId));
+            Assert.That(device.DeviceTag, Is.EqualTo(validDeviceTag));
+            Assert.That(device.Manufacturer, Is.EqualTo(validManufacturer));
+            Assert.That(device.ModelNum, Is.EqualTo(validModelNum));
+            Assert.That(device.SerialNum, Is.EqualTo(validSerialNum));
+            Assert.That(device.YearManufactured, Is.EqualTo(validYearManufactured));
+            Assert.That(device.Notes, Is.EqualTo(validNotes));
+
+
         }
     }
 }
--- a/BEIMA.Backend.Test/MongoService/BaseDeviceTest.cs
+++ b/BEIMA.Backend.Test/MongoService/BaseDeviceTest.cs
@@ -131,7 +131,119 @@ namespace BEIMA.Backend.MongoService.Test
             Assert.That(device.YearManufactured, Is.EqualTo(validYearManufactured));
             Assert.That(device.Notes, Is.EqualTo(validNotes));
 
+            device.SetLastModified(validDate, validUser);
+            var lastModified = device.LastModified;
+            Assert.That((DateTime)lastModified.GetElement("date").Value, Is.EqualTo(validDate).Within(5).Seconds);
+            Assert.That((string)lastModified.GetElement("user").Value, Is.EqualTo(validUser));
 
+            device.SetLocation(validBuildingId, validLocationNotes, validLatitude, validLongitude);
+            var location = device.Location;
+            Assert.That((ObjectId)location.GetElement("buildingId").Value, Is.EqualTo(validBuildingId));
+            Assert.That((string)location.GetElement("notes").Value, Is.EqualTo(validLocationNotes));
+            Assert.That((string)location.GetElement("latitude").Value, Is.EqualTo(validLatitude));
+            Assert.That((string)location.GetElement("longitude").Value, Is.EqualTo(validLongitude));
+
+            device.AddField(validKey1, validValue1);
+            device.AddField(validKey2, validValue2);
+            device.AddField(validKey3, validValue3);
+            var fields = device.Fields;
+            Assert.That((string)fields.GetElement(validKey1).Value, Is.EqualTo(validValue1));
+            Assert.That((bool)fields.GetElement(validKey2).Value, Is.EqualTo(validValue2));
+            Assert.That((int)fields.GetElement(validKey3).Value, Is.EqualTo(validValue3));
+        }
+
+        [Test]
+        public void BaseDeviceNotInstantiated_InstantiateUsingObjectInitializerWithAllNullStringValues_GetBsonDocumentThrowsException()
+        {
+            var device = new BaseDevice
+            {
+                Id = validObjId,
+                DeviceTypeId = validDeviceTypeId,
+                DeviceTag = null,
+                Manufacturer = null,
+                ModelNum = null,
+                SerialNum = null,
+                YearManufactured = validYearManufactured,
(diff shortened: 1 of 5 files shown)


BEIMA.DB.Schemas/device-schema.txt
```diff
@@ -1,6 +1,5 @@
 {
     "deviceTypeId": String,
-    "serialNumber": String,
```
josephmoore99, line 3: Why are we removing this?

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// Full constructor to instantiate a Device object.
+        /// </summary>
+        /// <param name="id">ObjectId of the object itself.</param>
+        /// <param name="deviceTypeId">ObjectId of the DeviceType this Device is linked to.</param>
+        /// <param name="deviceTag"></param>
+        /// <param name="manufacturer"></param>
+        /// <param name="modelNum"></param>
+        /// <param name="serialNum"></param>
+        /// <param name="yearManufactured"></param>
+        /// <param name="notes"></param>
```
josephmoore99, line 34: params

BEIMA.DB.Schemas/device-schema.txt
roguedeleter: "serialNumber" is a duplicate value, as "serialNum" already exists.

BEIMA.Backend/MongoService/BaseDevice.cs
roguedeleter: I thought these were self explanatory, what type of param comments need to be put here?

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// <summary>
+        /// Gets a BsonDocument that represents the current state of the BaseDevice object.
+        /// </summary>
+        /// <returns>BsonDocument that represents the current state of the BaseDevice object. Conforms to the schema located in BEIMA.DB.Schemas.</returns>
+        /// <exception cref="ArgumentNullException">Throws exception when any of the required fields are null.</exception>
+        public BsonDocument GetBsonDocument()
+        {
+            if (DeviceTag == null)
+            {
+                throw new ArgumentNullException($"BaseDevice - {nameof(DeviceTag)} is null");
```
josephmoore99, line 95: Not sure how I feel about throwing an exception here on null data, especially for things like Notes or LastModified that may not necessarily be required have any data in them. If we want to throw an exception, I think it would be better as a NullReferenceException instead of ArgumentNullException?

AshlynAdamson, line 95: I think we can put this null check into it's own method, and pass in the field as the argument. That way we have multiple method calls, rather than almost duplicate if and throw statements

josephmoore99: I think we should only throw exceptions on data we think should absolutely be required.

roguedeleter: Null data indicates that the programmer did not pass anything in for that field. The programmer needs to pass in an empty string, or some other default value. If null data has been passed in, it means something has gone wrong and an exception must be thrown.

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+            }
+            return this.ToBsonDocument();
+        }
+
+        /// <summary>
+        /// Adds a custom field to the "fields" object.
+        /// </summary>
+        /// <param name="key">The ObjectId of the linked field in the linked DeviceType.</param>
+        /// <param name="value">The actual value of the field.</param>
+        public void AddField(string key, dynamic value)
```
AshlynAdamson, line 133: This method adds a value for a custom field that was defined in the device type. Do we need a method to do that, but for a custom field on a single device? I think that was a scenario that Brian had described. Maybe this will be tackled in a future task, but thought I'd bring it up to make sure

BEIMA.Backend/MongoService/BaseDevice.cs
roguedeleter: Comparing the descriptions of both exceptions, it seemed like the wording of ArgumentNullException seems more appropriate to me.
https://docs.microsoft.com/en-us/dotnet/api/system.nullreferenceexception?view=net-6.0
https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception?view=net-6.0

It seems like a NullReferenceException is thrown when "there is an attempt to dereference a null object reference."

Versus, a ArgumentNullException is thrown when "a null reference is passed to a method that does not accept it as a valid argument." This seems more appropriate, since the only way those values can be null, is if a null was passed in as an argument to the constructor/object initializer.

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// </summary>
+        /// <param name="id">ObjectId of the object itself.</param>
+        /// <param name="deviceTypeId">ObjectId of the DeviceType this Device is linked to.</param>
+        /// <param name="deviceTag"></param>
+        /// <param name="manufacturer"></param>
+        /// <param name="modelNum"></param>
+        /// <param name="serialNum"></param>
+        /// <param name="yearManufactured"></param>
+        /// <param name="notes"></param>
+        public BaseDevice(ObjectId id, ObjectId deviceTypeId, string deviceTag, string manufacturer, string modelNum, string serialNum, int yearManufactured, string notes)
```
josephmoore99, line 35:
```suggestion
        public BaseDevice(ObjectId id, ObjectId deviceTypeId, int yearManufactured, string deviceTag = string.Empty, string manufacturer = string.Empty, string modelNum = string.Empty, string serialNum = string.Empty, string notes = string.Empty)
```
We can set defaults for constructor parameters, also should probably split this into a couple lines.

AshlynAdamson (changes requested):
Left some feedback in other comments. Looks great!

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+        {
+            //BsonClassMap.RegisterClassMap<BaseDevice>();
+
+            Id = id;
+            DeviceTypeId = deviceTypeId;
+            DeviceTag = deviceTag ?? string.Empty;
+            Manufacturer = manufacturer ?? string.Empty;
+            ModelNum = modelNum ?? string.Empty;
+            SerialNum = serialNum ?? string.Empty;
+            YearManufactured = yearManufactured;
```
josephmoore99, line 45: Should we make this a nullable int? Doesn't seem like we should be requiring every device to have this instantiated.

BEIMA.Backend.Test/MongoService/BaseDeviceTest.cs
```diff
@@ -0,0 +1,207 @@
...
+            Assert.That(device.Id, Is.EqualTo(validObjId));
+            Assert.That(device.DeviceTypeId, Is.EqualTo(validDeviceTypeId));
+            Assert.That(device.DeviceTag, Is.EqualTo(validDeviceTag));
+            Assert.That(device.Manufacturer, Is.EqualTo(validManufacturer));
+            Assert.That(device.ModelNum, Is.EqualTo(validModelNum));
+            Assert.That(device.SerialNum, Is.EqualTo(validSerialNum));
+            Assert.That(device.YearManufactured, Is.EqualTo(validYearManufactured));
+            Assert.That(device.Notes, Is.EqualTo(validNotes));
+
+            device.SetLastModified(validDate, validUser);
```
josephmoore99, line 48: I think it's good practice to set all of the object stuff at the beginning, and than do all the assertions at the end.

BEIMA.Backend.Test/MongoService/BaseDeviceTest.cs
```diff
@@ -0,0 +1,207 @@
...
+            device.AddField(validKey2, validValue2);
+            device.AddField(validKey3, validValue3);
+            var fields = device.Fields;
+            Assert.That((string)fields.GetElement(validKey1).Value, Is.EqualTo(validValue1));
+            Assert.That((bool)fields.GetElement(validKey2).Value, Is.EqualTo(validValue2));
+            Assert.That((int)fields.GetElement(validKey3).Value, Is.EqualTo(validValue3));
+        }
+
+        [Test]
+        public void BaseDeviceNotInstantiated_InstantiateUsingObjectInitializerWithAllNullStringValues_GetBsonDocumentThrowsException()
```
josephmoore99, line 114: If you want, you can set up various test cases using the same method. You would use something like:
`[TestCase(1, "2", null)]
[TestCase(3, "4", "5"]
public void MyTest(int number, string name, string value)`
Although, this would probably need a ridiculous number of method parameters .

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+            Notes = notes ?? string.Empty;
+            Fields = new BsonDocument();
+            Location = new BsonDocument();
+            LastModified = new BsonDocument();
```
josephmoore99, line 49: We can just call the other constructor for these.

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+            Fields.Set(key, value);
+        }
+
+        /// <summary>
+        /// Sets all of the fields in the "location" object of the BaseDevice.
+        /// </summary>
+        /// <param name="buildingId"></param>
+        /// <param name="notes"></param>
+        /// <param name="latitude"></param>
+        /// <param name="longitude"></param>
```
josephmoore99, line 144: params

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+        public void SetLocation(ObjectId buildingId, string notes, string latitude, string longitude)
```
josephmoore99, line 145:
```suggestion
        public void SetLocation(ObjectId buildingId, string notes = string.Empty, string latitude = string.Emtpy, string longitude = string.Empty)
```

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+            Location.Set("latitude", latitude);
+            Location.Set("longitude", longitude);
+        }
+
+        /// <summary>
+        /// Sets all of the fields in the "lastModified" object of the BaseDevice.
+        /// </summary>
+        /// <param name="date">Date of when the Device was last modified.</param>
+        /// <param name="user">Username of the user who last modified the Device.</param>
+        public void SetLastModified(DateTime? date, string user)
```
josephmoore99, line 163:
```suggestion
        public void SetLastModified(DateTime? date, string user = string.Empty)
```

josephmoore99 (changes requested):
Good stuff, will be nice to have some more well enforced typing for database objects.

BEIMA.Backend/MongoService/BaseDevice.cs
josephmoore99: I think you may still be able to manually set the parameters to null when you call it if you specifically specify it.

BEIMA.Backend/MongoService/BaseDevice.cs
josephmoore99: You can literally just copy the param name if it's simple enough. Like "The device tag" for "deviceTag".

BEIMA.Backend/MongoService/BaseDevice.cs
josephmoore99: I think that for now, this is fine, but in the future, I think we would likely want to use some sort of interaction with a deviceType object when setting these fields. Or even just query the database to see of the field value is valid for the given device type.

BEIMA.Backend.Test/MongoService/BaseDeviceTest.cs
Kenny-Miller: Would we want to just pull the assertions at the top out into their own test case?

josephmoore99: We shouldn't need to, setting the various properties should not be affecting one another.

AshlynAdamson: All the fields need to be set for a Bson to be returned correctly. There's another test case that only sets a few fields and leaves the rest null. Moving around the asserts is a good idea.

Kenny-Miller (approved):

BEIMA.Backend/MongoService/BaseDevice.cs
roguedeleter: Resolved [95e7484](https://github.com/cs481-ekh/s22-beima/pull/114/commits/95e7484f1c1e7fa2a0b2e1fa32c92df232cf3a0d)

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// <param name="deviceTypeId">ObjectId of the DeviceType this Device is linked to.</param>
+        /// <param name="deviceTag"></param>
+        /// <param name="manufacturer"></param>
+        /// <param name="modelNum"></param>
+        /// <param name="serialNum"></param>
+        /// <param name="yearManufactured"></param>
+        /// <param name="notes"></param>
+        public BaseDevice(ObjectId id, ObjectId deviceTypeId, string deviceTag, string manufacturer, string modelNum, string serialNum, int yearManufactured, string notes)
+        {
+            //BsonClassMap.RegisterClassMap<BaseDevice>();
```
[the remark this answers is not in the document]
roguedeleter: Resolved [95e7484](https://github.com/cs481-ekh/s22-beima/pull/114/commits/95e7484f1c1e7fa2a0b2e1fa32c92df232cf3a0d)

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,174 @@
...
+            YearManufactured = yearManufactured;
+            Notes = notes ?? string.Empty;
+            Fields = new BsonDocument();
+            Location = new BsonDocument();
+            LastModified = new BsonDocument();
+        }
+
+
+        [BsonId]
+        public ObjectId Id { get; set; }
```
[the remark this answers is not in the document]
roguedeleter: Fixed [6d85cb2](https://github.com/cs481-ekh/s22-beima/pull/114/commits/6d85cb251c5a87d0b5701a4653d6c9306c76de53)

BEIMA.Backend/MongoService/BaseDevice.cs
roguedeleter: @josephmoore99 @AshlynAdamson If the programmer can still technically pass in null (either unintentionally or intentionally), then setting defaults for constructor parameters is not a good fix. The ??= syntax I've used in this constructor will take care of both defaults and null values.

BEIMA.Backend/MongoService/BaseDevice.cs
roguedeleter: I'm not sure if MongoDB allows nullable ints, I will look into this.

BEIMA.Backend/MongoService/BaseDevice.cs
roguedeleter: The way I envisioned the full constructor being used, is that the programmer will set variables like so:
`
var id = ...
var deviceTypeId = ...
var deviceTag = ...
(and then all the other properties) 
`
Then, they will pass in all those parameters into the constructor:
`
var device = new BaseDevice(id, deviceTypeId, deviceTag, manufacturer, modelNum, serialNum, yearManufactured, notes);
`

What if in the first block of code, one of those parameters was set to null (because maybe the front end has a bug that made it send a null)? Then, my current implementation will handle this properly and set a default value.

josephmoore99: I think it's nice to have though because it means you aren't forced to specify each parameter every single time, especially if you don't have the associated values.

BEIMA.Backend/MongoService/BaseDevice.cs
josephmoore99: If anything, we could have it be set to -1 or something.

BEIMA.Backend/MongoService/BaseDevice.cs
roguedeleter: @josephmoore99 The programmer must specify each parameter each time. There should not be a case where we only specify some fields and not all. The endpoints we write will be generalized, and we must assume that all of the fields will be populated with some sort of data. If it turns out that one field is null, it is okay since the constructor will handle it properly.

josephmoore99: Alright, if that's the case we can leave it as is. It was more of a quality of life improvement.

roguedeleter: I definitely understand the point about quality of life, since typing all of those parameters is such a pain. However, in this case, we would have to sacrifice some code/type safety which I am not very comfortable with on such a critical object.

BEIMA.Backend/MongoService/BaseDevice.cs
roguedeleter: MongoDB can store null values, but there is no way to truly specify it as an int if it is null. I am just going to store -1 as the default value as an alternative.

BEIMA.Backend.Test/MongoService/BaseDeviceTest.cs
roguedeleter: Resolved [bbc776a](https://github.com/cs481-ekh/s22-beima/pull/114/commits/bbc776a0b050b6457993e35109a7704be0e6b827)

BEIMA.Backend/MongoService/BaseDevice.cs
roguedeleter: Resolved [401e45f](https://github.com/cs481-ekh/s22-beima/pull/114/commits/401e45f2842b9ab766d956b9ff5b098c2098b2a0)

BEIMA.Backend.Test/MongoService/BaseDeviceTest.cs
roguedeleter: Since I've already written it this way, and also the number of method parameters is so large, I will keep it like this. Going forward, I will try and make use of TestCases.

roguedeleter:

AshlynAdamson (approved):

BEIMA.Backend/MongoService/BaseDevice.cs
```diff
@@ -0,0 +1,168 @@
...
+            Location = new BsonDocument();
+            LastModified = new BsonDocument();
+        }
+
+        /// <summary>
+        /// Internal method to check if a particular property of this class is null
+        /// </summary>
+        /// <param name="arg">The Property being null checked.</param>
+        /// <param name="name">The name of the Property, to be returned in the exception message.</param>
+        /// <exception cref="ArgumentNullException"></exception>
```
josephmoore99, line 88: Forgot exception description.

roguedeleter: Resolved [7c9c90e](https://github.com/cs481-ekh/s22-beima/pull/114/commits/7c9c90e00b2aae595377c7263457998fac01415b)

josephmoore99 (approved):

AshlynAdamson (approved):

Issue #79, opened 2022-02-17
Initial Device and Device Type Views


Kenny-Miller:
#1
### Help Page
As a user, I should be able to view a help page that contains instructions.

### Device Type List Page
As a user, I should be able to view a page that contains a list view of all the device type.
As a user, on the device type list view, I should be able to click a button to create a new device type entry 
(Routes to a Device Type Detail View)
As a user, if I click on one of the device type items, I should be routed to a device type view page.

### Device Type Detail View
As a user, I should be able to view a page that contains a detailed device type view.
As a user, on the device type list view, I should be able to click a button to update a device template entry 
As a user, on the device type list view, I should be able to click a button to edit a device template entry 

### Device Entry List Page
As a user, I should be able to view a page that contains a list view of all the device entry.
As a user, on the device entry list view, I should be able to click a button to create a new device entry 
(Routes to a Device Entry Detail View)
As a user, if I click on one of the device entry items, I should be routed to a detailed device entry view page.

### Device Entry Detail View
As a user, I should be able to view a page that contains a detailed device type view.
As a user, on the device type list view, I should be able to click a button to update a device template entry 
As a user, on the device type list view, I should be able to click a button to edit a device template entry 

**Acceptance Criteria:**
- [x] I should be able to view a page with a device type list.
- [x] I should be able to view a page with a detailed device type.
- [x] I should be able to view a page with a device entry list.
- [x] I should be able to view a page with a detailed device entry.

PR #126, opened 2022-02-25
Get Device API call

### LINKED ISSUE #94: JS get individual device entry connection
References user story #82

As a user, I need to be able to view a specific device entry so that I can see the data associated with a entry.

thessbsu:
Closes #94

This is from my apiCalls.js I used to calibrate the output, just shows usage
const api = require ('./ApiClient');
const returnVar = await api.GetDevice(&lt;device ID&gt;);

Running the api calls and the output. The actual return from the GetDevice call is the JSON object only. The parts after it are showing the members being accessed and what is there.

$ node apiCalls.js

const returnVar= await api.GetDevice(&lt;invalidObjectId&gt;);
{ status: '400', response: 'Invalid id.' }
returnVar.status: 400
returnVar.response: Invalid id.

const returnVar= await api.GetDevice(&lt;validDeviceIdNotInDb&gt;);
{ status: '404', response: 'Device could not be found.' }
returnVar.status: 404
returnVar.response: Device could not be found.

const returnVar= await api.GetDevice(&lt;deviceIdThatCurrentlyExistsInDb&gt;);
{
  status: '200',
  response: [
    { name: '_id', value: '620aeb23f50067dd0535bab3' },
    { name: 'deviceTypeId', value: 'a' },
    { name: 'serialNumber', value: 'a12345' },
    { name: 'updatedDeviceField', value: '123' }
  ]
}
returnVar.status: 200
returnVar.response[0].name: _id
returnVar.response[0].value: 620aeb23f50067dd0535bab3
etc..


josephmoore99:
I don't think this will affect what you have here, but I have updated the format for the json response for the GET endpoint. It should now be in a much more jsony format. It should just have the properties and their values now rather than having the "name" "value" tuples.

BEIMA.Client/services/ApiClient.js
```diff
@@ -0,0 +1,32 @@
+//pull in the vars file to get the API URL
+require('dotenv').config({ path: '../.env' })
+
+//axios package includes it's own try catch and error output functions 
+
+const axios = require('axios');
+
+const API_URL = process.env.REACT_APP_API_URL;
+
+/// <summary>
```
josephmoore99, line 10: I know this is our header formatting for our C# code, but do we want to use the same formatting here?

BEIMA.Client/services/ApiClient.js
```diff
@@ -0,0 +1,32 @@
...
+      }
+  });
+  
+  return response;
+}
+
+//module.exports.<Custom Name> = <Function Name>
+//OR
+//module.exports.<Function Name> = <Function Name>
+module.exports.GetDevice = GetDevice;
```
AshlynAdamson, line 32: I think `export default GetDevice` will work here too. That's what we've been following for the other JS code.

AshlynAdamson:
I had envisioned these JS connection code tasks to have a bit more with them. The idea was to get a static UI up, and then these tasks would make functions to call the endpoint, like you've done here, but also use them in the UI, making that connection from the backend to the frontend. That is why I have mentioned before that I think these tasks needed to wait for the UI to be done.

Maybe I am the only one that interpreted it this way during sprint planning, and if that's the case then we can role with this and we will have to make another story for next sprint that updates the UI with these calls since story #79 is for initial pages with no functionality to them yet.

BEIMA.Client/services/ApiClient.js
thessbsu: Would that interfere with other functions being included in the file?

thessbsu:
> I don't think this will affect what you have here, but I have updated the format for the json response for the GET endpoint. It should now be in a much more jsony format. It should just have the properties and their values now rather than having the "name" "value" tuples.

It shouldn't affect it, I just rewrap it with the return code. The only change I can see is not having to use array indexing to access the members since they'll be flat.

BEIMA.Client/services/ApiClient.js
AshlynAdamson: There's currently only one function in this file, and it might be a good idea to keep each API call in its own file, like how the backend is doing. But if we want to include all the functions in one file, then you can add `export default` in front of the function declaration

thessbsu:
> 

I think I get what you're saying.
The API client should return a React component to be rendered?

I was thinking "this needs to get the data so the calling page can add it to the fields" rather than "return a render-able piece of a page"

If it needs to return a React component, I can do that, it will take a bit of reading and experimenting to get it right.

AshlynAdamson:
> I think I get what you're saying. The API client should return a React component to be rendered?
> 
> I was thinking "this needs to get the data so the calling page can add it to the fields" rather than "return a render-able piece of a page"
> 
> If it needs to return a React component, I can do that, it will take a bit of reading and experimenting to get it right.
> 
> If the calling page will be adding data to the fields I have the code for using them outside the module that should be usable in a page. The api client would need to be in a requires directive then, inside a function in the page would have:
> 
> const retVar = await api.GetDevice(<param>);
> 
> then the data itself is in the retVar to be assigned to the fields

I don't think we need to return a render React item, returning the data how you have it now should be fine. What I'm referring to is updating the calling page to actually call this function and verify the page can is updated correctly.

thessbsu:
> I don't think we need to return a render React item, returning the data how you have it now should be fine. What I'm referring to is updating the calling page to actually call this function and verify the page is updated correctly.

When thinking it might need to be a component I searched around and found a use case for something else, but I thought in the React component it would look something like this:

import React from 'react';
const api = require(&lt;relative path to ApiClient&gt;);

function ReactFunction() {
  
  function FillDevice(deviceId) {
    const getMyData = await api.GetDevice(deviceId);
	
	// fill device fields or pass the data to another page etc
       "set visible field's value/text to" getMyData.id;
  }
  
  return (
    &lt;ul&gt;
        &lt;li&gt;
          &lt;button onClick={() =&gt; FillMyDeviceFields(deviceId)}&gt;
          &lt;/button&gt;
        &lt;/li&gt;
    &lt;/ul&gt;
  );
}

BEIMA.Client/services/ApiClient.js
Kenny-Miller: You can just put export infront of const GetDevice... if you are wanting to have multiple exported functions in the file. I think the module.exports you have also works just as well.

Kenny-Miller (changes requested):
I think it would be better if you just returned this

const res = {
   status: response.status,
   response: response.data
}
return res

BEIMA.Client/services/ApiClient.js
thessbsu: I tried a few different ways of exporting and a few complained when trying to call them externally, I'll try the export in front of the function name and see how it likes it, I can't remember if I tried that or not.

Kenny-Miller (changes requested):
If you await, you don't need to use the .then statement. You just would wrap it in a try catch to handle errors. Likewise using the .then you don't need the await. https://stackabuse.com/making-asynchronous-http-requests-in-javascript-with-axios/

AshlynAdamson:
> import React from 'react'; const api = require(<relative path to ApiClient>);
> 
> function ReactFunction() {
> 
> function FillDevice(deviceId) { const getMyData = await api.GetDevice(deviceId);
> 
> ```
> // fill device fields or pass the data to another page etc
>    "set visible field's value/text to" getMyData.id;
> ```
> 
> }
> 
This is what I'm referring to. The pages are currently being developed with static device fields, so they already have rendered components for that. They would just need to include a function to dynamically populate the device fields using the API call function you've created, instead of using a static list of data.

Kenny-Miller:
> Currently, the pages don't know what those API functions are, that's why I think adding it to the pages needs to be done when we create the API function, therefore we should wait for the UI to be done. This PR can stay up, but once we get the UI done we should update the PR to incorporate the API calls into the page. Maybe we should have a discussion this evening about this.

Its fine doing both at the same time and merging the pr's in so long as we through in a task to merge them together. The ui should be using mock data to base itself off anyways while creating the view. That way the switch over will be as easy as just pasting a function call.

thessbsu:
> If you await, you don't need to use the .then statement. You just would wrap it in a try catch to handle errors. Likewise using the .then you don't need the await. https://stackabuse.com/making-asynchronous-http-requests-in-javascript-with-axios/

OH! Yes, it's likely still there as a remnant from when I did do some actions in the Then portion.

AshlynAdamson:
> Its fine doing both at the same time and merging the pr's in so long as we through in a task to merge them together. The ui should be using mock data to base itself off anyways while creating the view. That way the switch over will be as easy as just pasting a function call. That's not to say the objects won't change, but that more so depends on the backend endpoints and what they return.

The scope of the UI tasks is to create initial pages, the actual functionality of it is out of scope for those tasks. We have quite a bit of WIP and you and I have been working on those tasks for longer than 1-2 working sessions. We should get the initial UI tasks closed first, then move on to adding the API calls to the pages as we get those functions created. 

If we wait and do them at the same time, it adds way more decencies on those tasks than we had originally had, which is not good. It will take longer for them to close and we need to keep tasks moving through the pipelines rather than having them sit in "In Progress" for pretty much the whole sprint.

BEIMA.Client/services/ApiClient.js
thessbsu: All the examples I see after looking to see if there's a standard appear to be using the JSDoc format.
https://en.wikipedia.org/wiki/JSDoc

I'll swap it for that since it seems to be nearly ubiquitous

BEIMA.Client/services/ApiClient.js
thessbsu: One per file would limit the amount of external code that gets loaded per page, and the "show one device" doesn't need to know about device/device type list functionality and vice versa (or any other operation).

I'll swap it tonight

josephmoore99 (dismissed):

Kenny-Miller (dismissed):

thessbsu:
doc style was updated to JSDoc format
renamed to indicate single use file for each function that will be implemented
moved to the src/services directory so the pages can access it

The axios catch block was readded since internally axios attaches the DB response and makes it available to be in the returned response, otherwise the return is:
Error: Request failed with status code 404
    at createError (createError.js:16:1)
    at settle (settle.js:17:1)
    at XMLHttpRequest.onloadend (xhr.js:54:1)

without the await the browser catches the error response and puts this in the console and doesn't even get back to my code
GET http://localhost:7071/api/device/621b99ef97238d8b7acdf40e 404 (Not Found)

BEIMA.Client/services/ApiClient.js
```diff
@@ -0,0 +1,24 @@
+import axios from 'axios';
```
josephmoore99, line 1: It looks like you have the same code in three separate files, is this intentional?

thessbsu: It might not be picking up that the original "apiclient" file was renamed to "GetDevice", and then "GetDevice" was moved from /services to /src/services.  The remove of apiclient and move of getdevice might not have been added to the commit. I'll have to check once I'm off work.

thessbsu: Just deleted them through the web interface, one file, where it's supposed to be, is included

josephmoore99 (dismissed):

josephmoore99 (dismissed):

roguedeleter (approved):

Kenny-Miller (approved):

PR #134, opened 2022-02-27
Implemented DeviceType Functionality in MongoConnector

### LINKED ISSUE #81: Implement Device Type Functionality in MongoConnector
#2 

This story will implement CRUD operations for device types in MongoDB. This is very similar to the operations implemented in #23.

- [x] #128
- [x] #129
- [x] #130
- [x] #131
- [x] #132
- [x] #133

### LINKED ISSUE #128: Implement Insert operation
#81 

Implement Insert operation for DeviceType in MongoConnector.

### LINKED ISSUE #129: Implement Get one device type operation
#81

Implement Get One operation for DeviceType in MongoConnector.

### LINKED ISSUE #130: Implement Get all devices type operation
#81

Implement Get All operation for DeviceType in MongoConnector.

### LINKED ISSUE #131: Implement Update one device type operation
#81

Implement Update operation for DeviceType in MongoConnector.

### LINKED ISSUE #132: Implement Delete device type operation
#81

Implement Delete operation for DeviceType in MongoConnector.

### LINKED ISSUE #133: Add DeviceTypeCollectionName env variable to build files
#81 

The DeviceTypeCollectionName environment variable is new, the build files need to be updated accordingly.

roguedeleter:
Closes #81 - story
Closes #128 
Closes #129 
Closes #130 
Closes #131 
Closes #132 
Closes #132 
Closes #133

This PR implements DeviceType functionality in MongoConnector. It also implements the related unit tests, as well as the new environment variables in beima.yml. The code was mostly the same as the existing code for Device functionality, so it should look pretty similar.

There was also a change to the DB schemas. I removed the "deviceTypeId" and "buildingId" fields from the DeviceType and Building schemas respectively, since that would actually just be the "_id" field that is automatically generated when inserted into the database. I added the "_id" field to all of the schemas since that is what we will expect to see stored in the DB.

### DIFF
--- a/BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
+++ b/BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
@@ -160,5 +160,121 @@ namespace BEIMA.Backend.Test.MongoService
             var result = mongo.IsConnected();
             Assert.IsTrue(result);
         }
+
+        //DeviceType tests
+
+        [Test]
+        public void ConnectorCreated_GetDeviceTypeGivenValidDeviceTypeId_DeviceDocumentReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            //Setup (inserting a document)
+            var doc = new BsonDocument {
+                { "name", "TestDeviceType"},
+                { "description", "This is a test" }
+            };
+            var insertResult = mongo.InsertDeviceType(doc);
+            Assert.IsNotNull(insertResult);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
+
+            //Test (retrieve the document)
+            var retrievedDoc = mongo.GetDeviceType((ObjectId)doc["_id"]);
+            Assert.IsNotNull(retrievedDoc);
+            Assert.That(retrievedDoc, Is.TypeOf(typeof(BsonDocument)));
+        }
+
+        [Test]
+        public void ConnectorCreated_GetAllDeviceTypes_DeviceDocumentListReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            //Setup (inserting a document)
+            var doc = new BsonDocument {
+                { "name", "TestDeviceType"},
+                { "description", "This is a test" }
+            };
+            var insertResult = mongo.InsertDeviceType(doc);
+            Assert.IsNotNull(insertResult);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
+
+            //Test (retrieve all documents)
+            var retrievedDocs = mongo.GetAllDeviceTypes();
+            Assert.IsNotNull(retrievedDocs);
+            Assert.That(retrievedDocs, Is.TypeOf(typeof(List<BsonDocument>)));
+        }
+
+        [Test]
+        public void DeviceTypeNotInserted_InsertDeviceTypeAndDeleteDeviceType_DeviceTypeHasBeenDeleted()
+        {
+            var mongo = MongoConnector.Instance;
+            var doc = new BsonDocument {
+                { "name", "TestDeviceType"},
+                { "description", "This is a test" }
+            };
+            //Insert device type
+            var insertResult = mongo.InsertDeviceType(doc);
+            Assert.IsNotNull(insertResult);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
+
+            //Delete device type
+            bool deleteResult = false;
+            if (insertResult != null)
+            {
+                deleteResult = mongo.DeleteDeviceType((ObjectId)insertResult);
+                Assert.IsNull(mongo.GetDeviceType((ObjectId)insertResult));
+            }
+            Assert.IsTrue(deleteResult);
+        }
+
+        [Test]
+        public void DeviceTypeNotInserted_InsertDeviceTypeAndUpdateDeviceType_DeviceTypeHasBeenUpdated()
+        {
+            var mongo = MongoConnector.Instance;
+            var doc = new BsonDocument {
+                { "name", "TestDeviceType"},
+                { "description", "This is a test" }
+            };
+            //Insert device type
+            var insertResult = mongo.InsertDeviceType(doc);
+            Assert.IsNotNull(insertResult);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
+
+            //Update device type
+            doc.AddRange(new BsonDocument { { "updatedDeviceTypeField", "123" } });
+            var updateResult = mongo.UpdateDeviceType(doc);
+            Assert.IsNotNull(updateResult);
+        }
+
+        [Test]
+        public void ConnectorCreated_GetDeviceTypeGivenInvalidDeviceTypeId_NullReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            //This is a valid ObjectId, but this is not in the database
+            var doc = mongo.GetDeviceType(ObjectId.GenerateNewId());
+            Assert.IsNull(doc);
+        }
+
+        [Test]
+        public void ConnectorCreated_DeleteInvalidDeviceType_FalseReturned()
+        {
+            var mongo = MongoConnector.Instance;
+            //This is a valid ObjectId, but this is not in the database
+            var result = mongo.DeleteDeviceType(ObjectId.GenerateNewId());
+            Assert.IsFalse(result);
+        }
+
+        [Test]
+        public void ConnectorCreated_InsertNullDeviceType_NullReturned()
+        {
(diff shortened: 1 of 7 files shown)


BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
```diff
@@ -160,5 +160,121 @@ public void ConnectorCreated_CheckIsConnected_TrueReturned()
             var result = mongo.IsConnected();
             Assert.IsTrue(result);
         }
+
+        //DeviceType tests
```
josephmoore99, line 164: I think you might find using regions to be pretty handy here:
https://www.c-sharpcorner.com/article/using-regions-to-improve-code-readability/

BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -188,5 +189,134 @@ public BsonDocument UpdateDevice(BsonDocument doc)
...
+                return null;
+            }
+        }
+
+        /// <summary>
+        /// Deletes from the "deviceTypes" collection, given the objectID.
+        /// </summary>
+        /// <param name="objectId">Corresponds to the "_id" field for a given document inside of MongoDB</param>
+        /// <returns>true if successful, false if not successful</returns>
+        public bool DeleteDeviceType(ObjectId objectId)
```
josephmoore99, line 269: Do we want to add a check to make sure that there aren't any devices associated with the given device type? Or do we want to do that somewhere else?

BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
roguedeleter: Oops, I was meaning to use regions there and was planning on going back to change that. Thanks for catching that, I'll fix that and add another region for Device tests.

BEIMA.Backend/MongoService/MongoConnector.cs
AshlynAdamson: I think a check would be good, then something can be passed up to frontend and a warning message can be displayed before the commit to deleting it. I think that check should be done else where though, we don't want to put too much responsibility into the Mongo Connector code

BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
```diff
@@ -160,5 +160,121 @@ public void ConnectorCreated_CheckIsConnected_TrueReturned()
...
+                { "description", "This is a test" }
+            };
+            var insertResult = mongo.InsertDeviceType(doc);
+            Assert.IsNotNull(insertResult);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
+
+            //Test (retrieve the document)
+            var retrievedDoc = mongo.GetDeviceType((ObjectId)doc["_id"]);
+            Assert.IsNotNull(retrievedDoc);
+            Assert.That(retrievedDoc, Is.TypeOf(typeof(BsonDocument)));
```
AshlynAdamson, line 182: We should have the asserts all at the end

BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
```diff
@@ -160,5 +160,121 @@ public void ConnectorCreated_CheckIsConnected_TrueReturned()
...
+        {
+            var mongo = MongoConnector.Instance;
+            var doc = new BsonDocument {
+                { "name", "TestDeviceType"},
+                { "description", "This is a test" }
+            };
+            //Insert device type
+            var insertResult = mongo.InsertDeviceType(doc);
+            Assert.IsNotNull(insertResult);
+            Assert.That(insertResult, Is.TypeOf(typeof(ObjectId)));
```
AshlynAdamson, line 215: Since this test is testing the delete functionality, we don't need asserts on if the inserting the device type was successful

BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
josephmoore99: You can use Assume instead of Assert for the first two asserts to accomplish this.

BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
josephmoore99: Same here, you can use Assume instead of Assert for semantic purposes.

BEIMA.DB.Schemas/building-schema.txt
```diff
@@ -1,5 +1,5 @@
 {
-    "buildingId": String,
+    "_id": String,
```
josephmoore99, line 2: Is this necessary? I'm concerned that this may cause conflicts with the actual building object, and it isn't immediately obvious to a new developer what this id is actually referencing. Is this a standard mongo thing?

josephmoore99: We would need to alter Device.cs to reflect this change as well.

roguedeleter: From the original PR comment:
> There was also a change to the DB schemas. I removed the "deviceTypeId" and "buildingId" fields from the DeviceType and Building schemas respectively, since that would actually just be the "_id" field that is automatically generated when inserted into the database. I added the "_id" field to all of the schemas since that is what we will expect to see stored in the DB.

buildingId and deviceTypeId are referring to the _id fields in the actual DeviceType and Building documents in Mongo. It does not make sense to have a buildingId field in the Building object when the _id field already exists and serves the same purpose. This is the same for deviceTypeId.

The only time we use the keys "buildingId" and "deviceTypeId" is when we are referencing the Building and DeviceType documents from another collection. In essence, a Device will keep using "buildingId" and "deviceTypeId" to reference the _id fields of the respective Building and DeviceType document.

Yes, this is standard Mongo behavior, "_id" is automatically inserted for any object and it can be used as a unique surrogate key for any document we store in Mongo.

The Device.cs class does not need to be changed. The class was already built to use the _id field.

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: I think the check should be placed at a higher level as well. When the user wants to delete a device type, then we should pop up a message on the front end showing how many devices they are about to delete. Only after they confirm, should we execute this function.

If we threw an exception at this low of a level, I would imagine that this function would almost always throw an exception. Probably better to do this at the endpoint level.

BEIMA.DB.Schemas/building-schema.txt
josephmoore99: I totally misinterpreted all of this and forgot to read your PR comment. I thought we were removing the buildingId attribute from the device schema, not the building schema. This makes much more sense now. Ignore my comment, sorry for the confusion lol.

BEIMA.Backend/MongoService/MongoConnector.cs
josephmoore99: Thinking about it, we may even need to do this at the frontend level, since the endpoint would just keep throwing an exception as well.

BEIMA.DB.Schemas/building-schema.txt
roguedeleter: No worries, it happens to all of us lol.

BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
roguedeleter: Fixed [bb6c78f](https://github.com/cs481-ekh/s22-beima/pull/134/commits/bb6c78f547b5e677d7518b3bf4db83c584eb42d7)

BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
roguedeleter: Fixed [bb6c78f](https://github.com/cs481-ekh/s22-beima/pull/134/commits/bb6c78f547b5e677d7518b3bf4db83c584eb42d7)

BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
roguedeleter: Fixed [bb6c78f](https://github.com/cs481-ekh/s22-beima/pull/134/commits/bb6c78f547b5e677d7518b3bf4db83c584eb42d7)

roguedeleter:
All outstanding issues have been resolved

josephmoore99 (dismissed):

AshlynAdamson (approved):

thessbsu (approved):

Kenny-Miller (approved):

PR #135, opened 2022-02-27
Adding pages for Add Device and Add Device Type

### LINKED ISSUE #121: Create Add Device view
References user story #79

We need a page to be able to add a device.

### LINKED ISSUE #122: Create Add Device Type view
References user story #79 

We need a page to be able to add a device type

AshlynAdamson:
Closes #121 
Closes #122 

Here are the pages for adding a new device and a device type. You can see the JSON and files uploaded printed out to the console when the add device/type is clicked. 

https://user-images.githubusercontent.com/58532471/155863863-72285d3a-e9bd-4f12-a3b3-bd00eb1767ec.mp4

https://user-images.githubusercontent.com/58532471/155863866-1c68a5f2-16bb-40f5-8b48-8aa10564d236.mp4


BEIMA.Client/src/App.js
```diff
@@ -1,23 +1,30 @@
...
   return (
     <BrowserRouter>
       <Routes>
         <Route path="/" element={<AppLayout />}>
           <Route index element={<HomePage />} />
+
+          <Route path="addDevice" element={<AddDevicePage/>}/>
+          <Route path="addDeviceType" element={<AddDeviceTypePage/>}/>
+
           <Route path="Help" element={<HelpPage />} />
```
josephmoore99, line 28: Is there a reason for the extra newlines here?

AshlynAdamson: Not really, got rid of them now

BEIMA.Client/cypress/integration/AddDeviceTypePage.spec.js
```diff
@@ -0,0 +1,90 @@
...
+})
+
+describe("Verify Data in fields is cleared when Add Device is selected", () => {
+  it('Enter data, click Add Device, verify fields are empty', () => {
+    cy.visit('http://localhost:3000/addDeviceType')
+    cy.get('#Name').type("new type")
+    cy.get('#Description').type("newly added type")
+    cy.get("[id='Device Type Notes']").type("meter from SEL")
+    cy.get("#addDeviceType").scrollIntoView().click()
+    cy.get('#Name').should('have.value', '')
```
josephmoore99, line 37: If we are going to get a field more than once, we should store it in a variable so that it doesn't have to search the DOM again. I think this will help keep the test time down. I might even suggest having some kinds of global variables for the most commonly used ones if that's a possibility.

BEIMA.Client/cypress/integration/AddDeviceTypePage.spec.js
```diff
@@ -0,0 +1,90 @@
...
+  })
+  it('Enter data into Device Type Notes field', () => {
+    cy.visit('http://localhost:3000/addDeviceType')
+    cy.get("[id='Device Type Notes']").type("meter from SEL")
+  })
+})
+
+describe("Verify Data in fields is cleared when Add Device is selected", () => {
+  it('Enter data, click Add Device, verify fields are empty', () => {
+    cy.visit('http://localhost:3000/addDeviceType')
```
josephmoore99, line 32: If it's not too complex, maybe we could find a way to keep the webpage loaded between various tests if they don't interfere with one another. Could possibly use some sort of setup function if cypress has that, idk. Mostly just trying to think of ways to keep the testing time down.

BEIMA.Client/src/pages/DeviceTypes/AddDeviceTypePage.js
```diff
@@ -0,0 +1,129 @@
...
+
+  // adds field to list
+  function addField(newField, event) {
+    let newList = deviceFields.concat(newField);
+    setDeviceFields(newList);
+    event.target.form.elements.newFieldForm.value = "";
+  }
+
+  // gathers input and puts it into JSON
+  function createJSON(event){
```
josephmoore99, line 53: Feels weird to use event here, maybe use something like "object"?

BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -0,0 +1,98 @@
...
+        setAdditionalDocs(formFields[i].files);
+        formFields[i].value = "";
+      }
+    }
+
+    setDeviceFields(defaultDeviceFields);
+    setFullDeviceJSON(fieldValues);
+    console.log(fullDeviceJSON);
+    console.log(deviceImage);
+    console.log(deviceAdditionalDocs);
```
josephmoore99, line 58: Leftover from debugging?

BEIMA.Client/cypress/integration/AddDeviceTypePage.spec.js
AshlynAdamson: Looking through the cypress docs, I didn't find an easy way for us to keep it open. We can't call `cy.visit()` outside of an `it` block

BEIMA.Client/cypress/integration/AddDeviceTypePage.spec.js
AshlynAdamson: We can't create variable like we are used to using cypress, there's a bit more involved with it. We can grab a DOM component, and have a `then` block following it and call commands on it in there. For places where that was appropriate I updated it, but for something like this where we go back and forth between components we can't use that `then` block.

AshlynAdamson:
> This looks really clean btw. Nice work!

Thank you! It was fun being able to dive into React. I definitely learned quite a bit about it here

BEIMA.Client/src/pages/Devices/AddDevicePage.js
AshlynAdamson: Yep, got rid of console statements

BEIMA.Client/src/pages/DeviceTypes/AddDeviceTypePage.js
AshlynAdamson: Updated to use "object"

BEIMA.Client/src/pages/Devices/AddDevicePage.js
AshlynAdamson: Actually, I need them there. They are printing the created JSONs and files uploaded, and besides printing them, they are not used and that causes the client build to fail. When we add the API calls to the pages that won't be a problem anymore

Kenny-Miller (dismissed):
Looks Great!

BEIMA.Client/src/pages/DeviceTypes/AddDeviceTypePage.js
josephmoore99: On further inspection, it looks like this is actually meant to be an event object, maybe we call it something like addButtonEvent instead.

AshlynAdamson: Updated to addButtonEvent

josephmoore99 (approved):

thessbsu (approved):

PR #136, opened 2022-02-27
Get device list

### LINKED ISSUE #99: Add device GET list endpoint
References user story #77

As a user, I need to be able to view a list of device entries so that I can see all devices that I have registered.

josephmoore99:
Closes #99 

Functionality for getting the list of all devices in the database.


BEIMA.Backend.FT/DeviceFT.cs
```diff
@@ -76,5 +79,98 @@ public async Task DeviceNotInDatabase_AddDevice_CreatesNewDevice()
...
+                        Latitude = "101.989",
+                        Longitude = "25.004",
+                        Notes = "Above",
+                    },
+                    Manufacturer = "Generic Company",
+                    ModelNum = "y5eyf",
+                    Notes = "qwerty",
+                    SerialNum = "t4vw",
+                    YearManufactured = 2012,
+                },
```
AshlynAdamson, line 139: Should we be inserting these into a test database instead of the actual devices database? I used Postman to hit this endpoint and I noticed it returned these same devices here

josephmoore99: The issue is that the functional tests are running externally from the backend application itself. So because the backend application uses the actual database when we start it up on the backend, the functional tests have to use that same one. 

This particular test is marked as explicit, which means it isn't going to run on the build machine, and can only be run manually.

AshlynAdamson: Oh I see, sounds good

AshlynAdamson (approved):

roguedeleter (approved):
Looks good!

PR #142, opened 2022-02-28
Device update endpoint

### LINKED ISSUE #100: Add device update endpoint
References user story #77

As a user, I need to be able to update a device entry, so that I can don't have to delete it if something changes.

josephmoore99:
Closes #100 

Implements device update endpoint.

### DIFF
--- a/BEIMA.Backend.FT/DeviceFT.cs
+++ b/BEIMA.Backend.FT/DeviceFT.cs
@@ -172,5 +172,56 @@ namespace BEIMA.Backend.FT
                 Assert.That(device.LastModified?.User, Is.EqualTo("Anonymous"));
             }
         }
+
+        [Test]
+        public async Task DeviceInDatabase_UpdateDevice_ReturnsUpdatedDevice()
+        {
+            // ARRANGE
+            var origDevice = new Device
+            {
+                DeviceTag = "E-6",
+                DeviceTypeId = "abc830495728312323103456",
+                Location = new Location
+                {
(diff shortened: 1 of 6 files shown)


Kenny-Miller (commented):
I think we should deserialize into a model of the request instead of a dynamic object just to provide clarity on what we are expecting to be sent in the request

josephmoore99:
> I think we should deserialize into a model of the request instead of a dynamic object just to provide clarity on what we are expecting to be sent in the request

I attempted this with the add device endpoint but ran into issues since the Device object uses BSON types for things like Location, LastModified, and Fields. This causes problems for the json deserializer because it doesn't know how to covert these nested objects into BSON documents. Instead I am using the device constructors to populate these fields. The constructors themselves will be able to run checks to make sure the dynamic attributes being passed in are the type we expect.

BEIMA.Backend.FT/DeviceFT.cs
```diff
@@ -186,8 +186,7 @@ public async Task DevicesInDatabase_GetDeviceList_ReturnsDeviceList()
                 Assert.That(device.LastModified?.User, Is.EqualTo("Anonymous"));
             }
         }
-
-        [Test]
+
```
josephmoore99, line 189:
```suggestion
        [Test]
```
Left out test tag on merge, woops.

AshlynAdamson (approved):

thessbsu (approved):

PR #150, opened 2022-03-01
Device list JS API call

### LINKED ISSUE #93: JS device entries list connection
References user story #82

As a user, I need to be able to view device entries in a list view on the device entry list page.

AshlynAdamson:
Closes #93

Function that calls the devicelist endpoint. I filter the results to what we need, here a pic of what it looks like. Still need to figure out how to incorporate the process.env variable

image


BEIMA.Client/src/service/GetDeviceList.js
```diff
@@ -0,0 +1,37 @@
...
+      if (error.response) {
+        return error.response;
+    }
+  });
+
+  let filteredFields = deviceListCall.data.map((item) => {
+    return {
+      id: item._id,
+      deviceType: item.deviceTypeId,
+      buildingId: item.location.buildingId,
```
josephmoore99, line 20: I know there is other data such as location.latitude, location.longitude, and location.notes, as well as lastModified.user. Should we consider including some of these? Also, I think I should note that the post endpoint expects location and lastModified data to be nested objects. But I am guessing that this is more intended for use as some top level device data for populating the list row information.

AshlynAdamson: Yes, the function is intended to be called to populate the device list page, and I figured these fields were the most important. We can add more if we feel other fields are also important

josephmoore99 (approved):

thessbsu (approved):

AshlynAdamson:
Not sure why the minutes are in here, I just merged in main.

josephmoore99 (approved):

AshlynAdamson:
- [ ] fix my merging issues

josephmoore99 (approved):

thessbsu (approved):

roguedeleter (approved):

PR #198, opened 2022-03-04
Bug/devicetype input

### LINKED ISSUE #196: Device Type Additional Fields Input
References user story #87 

In the additional fields, after clicking edit, cannot type to update the text.

Kenny-Miller:
Closes #196

Fixes additional fields not being able to have their input changed when in edit mode.

### DIFF
--- a/BEIMA.Client/cypress/integration/DevicePage.spec.js
+++ b/BEIMA.Client/cypress/integration/DevicePage.spec.js
@@ -25,11 +25,11 @@ describe('Device Page', () => {
     cy.get('[id=deviceLatitude]').should('exist')
     cy.get('[id=deviceLongitude]').should('exist')
     cy.get('[id=locationNotes]').should('exist')
-    cy.get('[id=deviceTag]').should('exist')
-    cy.get('[id=deviceModelNumber]').should('exist')
-    cy.get('[id=deviceSerialNumber]').should('exist')
-    cy.get('[id=deviceManufacturer]').should('exist')
-    cy.get('[id=deviceYearManufactured]').should('exist')
+    cy.get('[id=fields]').within(() => {
+      cy.get('input').each((val, index, collection) => {
+        cy.wrap(val).should('exist')
+      })
+    })
 
     // disabled
     cy.get('[id=deviceNotes]').should('be.disabled')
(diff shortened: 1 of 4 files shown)


BEIMA.Client/cypress/integration/DevicePage.spec.js
```diff
@@ -25,23 +25,23 @@ describe('Device Page', () => {
     cy.get('[id=deviceLatitude]').should('exist')
     cy.get('[id=deviceLongitude]').should('exist')
     cy.get('[id=locationNotes]').should('exist')
-    cy.get('[id=deviceTag]').should('exist')
-    cy.get('[id=deviceModelNumber]').should('exist')
-    cy.get('[id=deviceSerialNumber]').should('exist')
-    cy.get('[id=deviceManufacturer]').should('exist')
```
josephmoore99, line 31: I am confused as to why we are removing these?

Kenny-Miller: They are picked up by this statement:

cy.get('[id=fields]').within(() => {
      cy.get('input').each((val, index, collection) => {
       // test stuff here
      })
    })

josephmoore99: Our fields attribute should only contain the attributes specific to the device type though right? Or is fields referring to something else? In that case wouldn't you also include removing the building stuff and deviceNotes?

BEIMA.Client/cypress/integration/DevicePage.spec.js
```diff
@@ -25,23 +25,23 @@ describe('Device Page', () => {
...
-    cy.get('[id=deviceYearManufactured]').should('exist')
+    cy.get('[id=fields]').within(() => {
+      cy.get('input').each((val, index, collection) => {
+        cy.wrap(val).should('exist')
```
AshlynAdamson, line 30: Doing it this way grabs the inputs on the page, and checks if it's there, dynamically. We aren't actually checking if the correct input field is there. I think how we had it before was good. We had actual expected results of what should be on the page

BEIMA.Client/cypress/integration/DevicePage.spec.js
Kenny-Miller: Its sort of combined. Fields contains those 4 ids I removes plus the attributes specific to the device types. The building stuff isn't included as I have it grouped differently in the UI. The id fields is mostly used so that I can grab the inputs from the custom fields as they will have dynamic keys.

BEIMA.Client/cypress/integration/DevicePage.spec.js
Kenny-Miller: Good point.  If those fields and custom fields don't get returned for some reason this test would still pass.

AshlynAdamson (approved):

josephmoore99 (approved):

PR #218, opened 2022-03-09
Add lat lon validation

### LINKED ISSUE #164: Invalid longitude/latitude error message
#157
We need to check the latitude/longitude the user inputs to ensure they are valid, and if not, display an error message and prevent the user from inserting/updating a device.

thessbsu:
Closes #164 

If an invalid entry is present it will be flagged when clicking "Add Device"
image

### DIFF
--- a/BEIMA.Client/src/pages/Devices/AddDevicePage.js
+++ b/BEIMA.Client/src/pages/Devices/AddDevicePage.js
@@ -26,7 +26,6 @@ const AddDevicePage = () => {
   const [deviceImage, setDeviceImage] = useState();
   const [deviceAdditionalDocs, setAdditionalDocs] = useState();
   const [fullDeviceJSON, setFullDeviceJSON] = useState({});
-  const [currentFormValues, setCurrentFormValues] = useState();  
   
   useEffect(() => {
     setPageName('Add Device')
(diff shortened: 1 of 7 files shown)


thessbsu:
fixed "merge into" branch

BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -19,28 +20,38 @@ const AddDevicePage = () => {
...
     let fieldValues = {};
+    let newErrors = {};
 
     for(let i = 0; i < formFields.length; i++){
       let formName = formFields[i].name;
       let fieldNames = Object.keys(deviceFields);
 
       if(fieldNames.includes(formName)){
         let formJSON =  {[formName] : formFields[i].value};
-        formFields[i].value = "";
```
AshlynAdamson, line 43: Can we keep this line, but move it to a better spot? I use it to clear the fields on a submit. Or maybe we can create a separate method that clears the fields after all the input validation is done and passes

thessbsu: I had tried to post a reply earlier but it doesn't look like it took.

Clearing the from will be in the Else statement when detecting errors. I'm hoping that setting the DB JSON object for the fields by selected device type to the currentDeviceFields variable will clear it. I've got some in progress stuff for that in another (unpublished at the moment) branch

AshlynAdamson (approved):
Looks good! Glad you got the formlist to work

josephmoore99 (approved):

roguedeleter (approved):

PR #219, opened 2022-03-10
Add device type page api

### LINKED ISSUE #169: Update Add Device Type to use API calling function
References user story #82 

AshlynAdamson:
Closes #169 

Added add device type api call to the add device type page. I also added some input validation for custom fields. This makes it so the user can't add empty fields or a field that is already added. 

https://user-images.githubusercontent.com/58532471/157730401-da13d013-0f68-42d1-afc9-26be5382f283.mp4


AshlynAdamson:
- [ ] discuss with team about left over device types after running locally
- [ ] based on discussion, decide on solution for removing left over device types

AshlynAdamson:
> * [x]  discuss with team about left over device types after running locally
> * [x]  based on discussion, decide on solution for removing left over device types

After discussion, this task #221 was created as the solution. It will be done in Sprint 3.

josephmoore99 (approved):

thessbsu (approved):

Issue #168, opened 2022-03-03
Add Device Page  - Clicking Add Device should add it to the DB


AshlynAdamson:
References user story #82 

We need to update the page to use the API function that calls the backend and ensure the data gets displayed correctly.

PR #222, opened 2022-03-10
Attach db call to dropdown

### LINKED ISSUE #220: Device Type DD needs to load from DB
References user story #82

When the page loads the Select Device Type dropdown needs to load the device types from the database and populate them as options to choose from

thessbsu:
Closes #220

Added a shared component to fill a dropdown.
It requires the items to fill with, the function to execute, and the style attribute for the DD to have as parameters.

The main page calls the GetDeviceTypes JS/API to get the fields to set in the DD


thessbsu:
- [ ] add test for making sure there's an item in the DD

BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -26,10 +28,32 @@ const AddDevicePage = () => {
...
+      let types = await getDeviceTypes();
+      setDeviceTypes(types);
+    }
+   loadData()
+  },[])
+  
+  const getDeviceTypes = async () => {
+    const deviceTypeData = await GetDeviceTypeList();
+    let data = []
+    for(let i = 0; i < deviceTypeData.response.length; i++){
```
AshlynAdamson, line 45: We should use the map function! I think it would look something like this:
`let data = deviceTypeData.response.map((item) => {
return { name: item.name, id : item.id}
});`

BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -26,10 +28,32 @@ const AddDevicePage = () => {
...
+      data.push({
+        name: deviceTypeData.response[i].name,
+        id: deviceTypeData.response[i].id
+      });
+    }
+    return data
+  }
+  
+  const getFieldsForTypeId = (deviceTypeId) => {
```
AshlynAdamson, line 54: My guess is that this will be completed in the PR for #168? Just wanted to check

thessbsu: either that one or another specifically targeted task for the work. at some point I'll edit that task to be a specific one like #220 (the task for this PR) is.

BEIMA.Client/src/shared/DropDown/FilledDropDown.js
```diff
@@ -0,0 +1,19 @@
+import { Dropdown } from 'react-bootstrap';
+
+const FilledDropDown = ({items, selectFunction, buttonStyle}) => {
```
AshlynAdamson, line 3: An idea that can happen later, but we can change the Building input to be a drop down and use this shared component to populate the available buildings. Nice work!

AshlynAdamson (changes requested):
I had some device types on my local DB and was able to see this in action. 
image
Looks good!

BEIMA.Client/src/shared/DropDown/FilledDropDown.js
thessbsu: All I did was rip off your idea, but thanks

BEIMA.Client/src/pages/Devices/AddDevicePage.js
thessbsu: - [ ] Let's map this

thessbsu: syntax was perfect Ashlyn

AshlynAdamson (approved):

josephmoore99 (approved):

PR #224, opened 2022-03-12
Update device type js call

### LINKED ISSUE #200: JS connection code for update Device Type
References user story #82

Need API call to update a Device Type

thessbsu:
Closes #200

confirmed it hits the correct API endpoint and delivers the payload to be parsed

### DIFF
--- a/BEIMA.Client/cypress/integration/App.spec.js
+++ b/BEIMA.Client/cypress/integration/App.spec.js
@@ -3,6 +3,6 @@
 describe('Open BEIMA Frontend', () => {
     it('Visits our running BEIMA Frontend', () => {
         cy.visit('http://localhost:3000')
-        cy.get('.App-link').should('have.attr', 'href', 'https://reactjs.org')
+        //cy.get('.App-link').should('have.attr', 'href', 'https://reactjs.org')
     })
 })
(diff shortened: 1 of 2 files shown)


thessbsu:
This branch was taken from feature/s2, changes 1 file by adding it.
feature/s2 passes
running it on just the branch passes
running it locally passes
running it on the PR fails
I have no idea why, it reports that it can't find the "Learn React" link on the default page??

AshlynAdamson:
> I have no idea why, it reports that it can't find the "Learn React" link on the default page??
There is a react tests test file that had one checked for the link on the starter react project page with the spinner. Since we aren't using the react testing framework at the moment, and no longer that the default spinner page, we can probably remove that test file. It should be under src/_tests? Let me double check and I'll update this comment.

AshlynAdamson (dismissed):

thessbsu:
I was thinking that was the source of the problem, but I'm not sure why it suddenly cropped up on this pull.

Kenny-Miller (approved):

josephmoore99:
image
It looks like it's failing because it's trying to test on the default react page that we got rid of? I could be totally wrong here.

thessbsu:
Wouldn't let a merge happen even after two approvals (because of test failures?) I commented out the check for the link but left the "go to the base site" part so that it confirms it opened at least, even if it gets redirected to another page when visiting the base url

PR #227, opened 2022-03-13
Add api calls to device type list page

### LINKED ISSUE #171: Update Device Type List page to use API calling function
References user story #82 

thessbsu:
Closes #171
Added items that are present on the pull to the existing fields
updated itemList.js to remove the placeholder text and take them as a parameter
added test to count the child elements of the list container to see if items are present when visiting the page


BEIMA.Client/src/pages/DeviceTypes/DeviceTypesPage.js
```diff
@@ -13,34 +14,19 @@ const DeviceTypesPage = () => {
...
-    let mapped = data.map(({deviceTypeId, name, description, numDevices}) => {
-      return {
-        id: deviceTypeId,
-        name: name,
-        description: description,
-        numDevices: numDevices
-      }
-    })
-    return mapped
+    let data = deviceTypeData.response.map((item) => { return { id: item.id, name: item.name, description: item.description, notes: item.notes, lastModified: item.lastModified, deviceTag: "item types don't have a tag", buildingName: "building name needed", numDevices: "count at DB or remove? counting length on the DT return gives count of device type"} });
```
josephmoore99, line 21: I'm not sure that we need the deviceTag attribute on this?

thessbsu: Same, the itemList.js that it's using is more appropriate for devices rather than device types. it might need some further revamping, or another file created.
Most of what was done here was proof it could use the API calls to get the data and display it once obtained.
I assume further refinement will happen to perfect it.

BEIMA.Client/src/pages/DeviceTypes/DeviceTypesPage.js
```diff
@@ -13,34 +14,19 @@ const DeviceTypesPage = () => {
...
-    let mapped = data.map(({deviceTypeId, name, description, numDevices}) => {
-      return {
-        id: deviceTypeId,
-        name: name,
-        description: description,
-        numDevices: numDevices
-      }
-    })
-    return mapped
+    let data = deviceTypeData.response.map((item) => { return { id: item.id, name: item.name, description: item.description, notes: item.notes, lastModified: item.lastModified, buildingName: "building name needed", numDevices: "count at DB or remove? counting length on the DT return gives count of device type"} });
```
AshlynAdamson, line 21: I think we can do without this filtering of the response since the GetDeviceTypeList already filters the data returned from the endpoint. We should be able to call the GetDeviceTypeList straight from the useEffect and get everything we need. It will need updated once we get the endpoint to return the count, but we can worry about that later.

BEIMA.Client/src/shared/ItemList/ItemList.js
```diff
@@ -48,7 +48,7 @@ const Item = ({item, RenderItem}) => {
   return (
     <div className={styles.item}>
       <div className={styles.row}> 
-        <div className={styles.itemName}>{item.deviceTag} - {"<Device Type Name>"} - {"<Building Name>"}</div>
+        <div className={styles.itemName}>{item.deviceTag} - {item.name} - {item.buildingName}</div>
```
AshlynAdamson, line 51: I forgot the device type list page also uses this, I don't think building info is needed for a device type. I'm thinking we could pass in another argument that indicates if it's a device or device type, then we can return a div with building info and then one without. We can use the ternary syntax, maybe something like `{device ? <div with building info> : <div without building info>}`

BEIMA.Client/src/pages/DeviceTypes/DeviceTypesPage.js
AshlynAdamson: > Same, the itemList.js that it's using is more appropriate for devices rather than device types. it might need some further revamping, or another file created. Most of what was done here was proof it could use the API calls to get the data and display it once obtained. I assume further refinement will happen to perfect it.

I left some comments regarding further refinement. I'm fine if we make another task to come back to that refinement and then get this PR through

BEIMA.Client/src/pages/DeviceTypes/DeviceTypesPage.js
thessbsu: I had it set that way to have something for the other fields, based on comments for the tag from Joseph and this one I've changed it to just return the response so any updates will be passed through

AshlynAdamson (approved):

josephmoore99 (approved):

PR #228, opened 2022-03-13
Feature/device storage

### LINKED ISSUE #15: Storage Provider
#2

Create a storage provider that can use either a local provider or an azure provider based on a configuration file. They will use a shared interface so that the actual process of performing crud operations are abstracted out. In addition the storage provider should make use of dependency injection to help facilitate this.

- [x] As a developer, based on a configuration file, I should be able to get or put a file in azure storage.
- [x] As a developer, based on a configuration file, I should be able to get or put a file on my system.
- [x] As a developer, I should have a service that I can use to manage interacting with storage.
- [x] Research how to upload a file and what endpoints to attach them to.
- [ ] Hook storage provider into existing endpoints

### LINKED ISSUE #232: Update AddDevice and UpdateDevice to accept a MIME request
References user story #15 

Add and update requests sent from the frontend will be a multipart form request. As such these endpoints will need to be configured to handle that request instead of a normal post request. The original post body will be serialized and sent in post.Form[data]

### LINKED ISSUE #233: Updated Device Endpoints to use Storage Provider
References user story #15

Currently the device endpoints do not make use of the storage provider. This task will add that functionality in to the add, update, get, and delete device endpoints.

Kenny-Miller:
Closes #15 
Closes #232 
Closes #233

The pull request closes the above tasks/stories. The main features of this pull request is to integrate the storage service in with the device endpoints so that files can be stored/accessed along side a device object. It also updated the add and update device endpoints to accept a mime request as the frontend will be sending a multipart form request in-order to send device json and files in one request.

Two other side changes were made to make this integration cleaner and to make the backend easier to work with. The first was a change to the device object. Instead of having the properties be a bson type (bson_element, bson_document, ect), the properties were changed to be represented by a c# object. For example device.Location would be a DeviceLocation object instead of a BsonDocument. The second change was the creation of c# objects to represent the incoming add device and update requests. This allows us to use a c# object instead of a dynamic one.


BEIMA.Backend.Test/DeviceFunctions/AddDeviceTest.cs
```diff
@@ -24,13 +30,41 @@ public async Task NoDevice_AddDevice_ReturnsValidId()
                   .Verifiable();
             MongoDefinition.MongoInstance = mockDb.Object;
 
+            // Setup storage provider.
+            var services = new ServiceCollection();
+            services.AddSingleton<IStorageProvider, AzureStorageProvider>();
+            var serviceProivder = services.BuildServiceProvider();
+            var storageProvider = serviceProivder.GetRequiredService<IStorageProvider>();
```
josephmoore99, line 37: Let's refactor this into a private method that returns the storage provider.

BEIMA.Backend/DeviceFunctions/AddDevice.cs
```diff
@@ -16,45 +17,80 @@ namespace BEIMA.Backend.DeviceFunctions
...
             }
             catch (Exception)
             {
                 return new BadRequestObjectResult(Resources.CouldNotParseBody);
             }
+
+            // Store attached files and add file uid to device
+            foreach (var file in reqForm.Files)
+            {
+                var fileUid = await _storage.PutFile(file);
```
josephmoore99, line 80: Is there any verification that needs to be done on files before putting them in?

BEIMA.Backend/DeviceFunctions/DeleteDevice.cs
```diff
@@ -35,6 +49,25 @@ public static class DeleteDevice
...
+                return new NotFoundObjectResult(Resources.DeviceNotFoundMessage);
+            }
+
+            var device = BsonSerializer.Deserialize<Device>(deviceDocument);
+
+            // Remove Files
+            foreach (var file in device.Files)
+            {
+                await _storage.DeleteFile(file.FileUid);
+                // Log to file if delete failed
```
josephmoore99, line 64:
```suggestion
                log.LogInformation($"Failed to delete a file: {file.FileUid.ToString()}.");
```

BEIMA.Backend/DeviceFunctions/UpdateDevice.cs
```diff
@@ -38,33 +55,85 @@ public static class UpdateDevice
...
+                buildingId,
+                data.Location.Notes,
+                data.Location.Latitude,
+                data.Location.Longitude
+            );
+            device.Fields = data.Fields;
+
+            // Check if there is a new photo, if so remove all others on the device
+            // Frontend only supports 1 photo, but a device can contain more
+            var updatePhoto = reqForm.Files.Any(file => file.Name == "photos");
```
josephmoore99, line 98: What if we want to delete the photo(s)?

BEIMA.Backend/DeviceFunctions/GetDevice.cs
```diff
@@ -22,40 +36,48 @@ public static class GetDevice
...
-                if (doc is null)
-                {
-                    return new NotFoundObjectResult(Resources.DeviceNotFoundMessage);
-                }
-
-                // Return the device.
-                var dotNetObj = BsonTypeMapper.MapToDotNetValue(doc);
-                return new OkObjectResult(dotNetObj);
+                var presignedUrl = await _storage.GetPresignedURL(photo.FileUid);
+                photo.Url = presignedUrl;
```
josephmoore99, line 69: So these aren't automatically populated on deserialization?

BEIMA.Backend/DeviceFunctions/UpdateDevice.cs
```diff
@@ -38,33 +55,85 @@ public static class UpdateDevice
...
+            if (updatePhoto && device.Photos.Count > 0)
```
josephmoore99, line 99: It feels weird to me to just delete all photos if there is any photos in the request. It seems like we should be able to individually select which photos to remove, add, or keep.

BEIMA.Backend/MongoService/Device.cs
```diff
@@ -152,10 +197,10 @@ public void SetLocation(ObjectId buildingId, string notes, string latitude, stri
             latitude ??= string.Empty;
             longitude ??= string.Empty;
 
-            Location.Set("buildingId", buildingId);
-            Location.Set("notes", notes);
-            Location.Set("latitude", latitude);
-            Location.Set("longitude", longitude);
+            Location.BuildingId = buildingId;
```
josephmoore99, line 200: I'm assuming that the .ToBsonDocument still works properly with this when sending it into the mongo methods?

BEIMA.Backend.Test/DeviceFunctions/AddDeviceTest.cs
```diff
@@ -24,13 +30,41 @@ public async Task NoDevice_AddDevice_ReturnsValidId()
...
+
             // Create request
-            var body = TestData._testDevice;
-            var request = CreateHttpRequest(RequestMethod.POST, body: body);
+            var data = new AddDeviceRequest()
```
josephmoore99, line 40: Let's put these objects to TestData.cs like we had with the json TestData, I think it just makes the tests a little more readable and allows us to reuse the data for multiple tests.

BEIMA.Backend.Test/RequestFactory.cs
```diff
@@ -44,5 +50,23 @@ public static HttpRequest CreateHttpRequest(RequestMethod requestMethod, Diction
...
+            DefaultHttpContext httpContext = new DefaultHttpContext();
+
+            var formKeys = new Dictionary<string, StringValues>();
+            formKeys.Add("data", data);
+
+            var formFiles = new FormFileCollection();
+            var fileStream = new ByteArrayContent(Encoding.ASCII.GetBytes("TestOne")).ReadAsStream();
+            var photoStream = new ByteArrayContent(Encoding.ASCII.GetBytes("TestTwo")).ReadAsStream();
+            formFiles.Add(new FormFile(fileStream, 0, fileStream.Length, "files", "file.txt"));
+            formFiles.Add(new FormFile(photoStream, 0, photoStream.Length, "photos", "photo.txt"));
```
josephmoore99, line 65: I think we should use the method arguments to populate the files and photos as opposed to hard coding them for every request that we test.

BEIMA.Backend.Test/MongoService/DeviceTest.cs
```diff
@@ -105,29 +74,29 @@ public void DeviceNotInstantiated_InstantiateUsingFullConstructor_DeviceInstanti
...
-            Assert.That((int)fields.GetElement(validKey3).Value, Is.EqualTo(validValue3));
+            Assert.That(fields[validKey1], Is.EqualTo(validValue1));
+            Assert.That(fields[validKey2], Is.EqualTo(validValue2));
+            Assert.That(fields[validKey3], Is.EqualTo(validValue3));
 
             var files = device.Files;
-            Assert.That(files.Contains(file1), Is.True);
-            Assert.That(files.Contains(file2), Is.True);
-            Assert.That(files.Contains(file3), Is.True);
+            Assert.That(files.Any(file => file.FileUid == validFileUid1 && file.FileName == validFileName1), Is.True);
```
josephmoore99, line 92:
```suggestion
            Assert.That(files.Single(file => file.FileUid == validFileUid1).FileName, Is.EqualTo(validFileName1));
```
I think I like doing it this way better, but that's just my personal preference.

BEIMA.Backend/DeviceFunctions/AddDevice.cs
```diff
@@ -16,45 +17,80 @@ namespace BEIMA.Backend.DeviceFunctions
...
-    public static class AddDevice
+    public class AddDevice
     {
+        private readonly IStorageProvider _storage;
+
+        /// <summary>
+        /// Constructor for the AddDevice Function
+        /// </summary>
+        /// <param name="storage">Depedency injected storage provider</param>
+        public AddDevice(IStorageProvider storage)
```
josephmoore99, line 28: So, I'm concerned that using this constructor is not going to play nicely with the current way we are using these endpoints. This seems to suggest that the request or the application itself should just know what storage provider to put in here on its own, which I don't think is the case. I think we would be better suited doing something like we are with the mongo connector, using instances from static or singleton classes.

BEIMA.Backend/DeviceFunctions/GetDevice.cs
Kenny-Miller: No since the urls will expire, we never store them.

BEIMA.Backend.Test/DeviceFunctions/AddDeviceTest.cs
```diff
@@ -24,13 +31,38 @@ public async Task NoDevice_AddDevice_ReturnsValidId()
                   .Verifiable();
             MongoDefinition.MongoInstance = mockDb.Object;
 
+            // Setup storage provider.
+            var storageProvider = StorageProviderExtensions.CreateAzureStorageProvider();
```
josephmoore99, line 35: Thinking about it, I actually think we should try and use mock storage providers for these unit tests. We should probably only have unit tests for the actual storage providers in the storage provider unit tests themselves.

BEIMA.Backend/DeviceFunctions/DeleteDevice.cs
Kenny-Miller: Wasn't sure if we wanted to use the logger or if we wanted to write all the failures to a file that someone could then open and read through.

BEIMA.Backend/DeviceFunctions/AddDevice.cs
Kenny-Miller: So it is singleton, at application startup the environment variables are read and a single storage provider is created just like the mongo connection. The only difference is that the application holds that reference and passes it to the endpoint rather then going to a class a getting the static instance held there.

BEIMA.Backend/DeviceFunctions/DeleteDevice.cs
thessbsu: a file sounds good so they can be reviewed, a "view logs" page can be added to the suggested improvements too
alternatively we could log to the DB

BEIMA.Backend/DeviceFunctions/AddDevice.cs
Kenny-Miller: Added making sure they have the right name and that the file doesn't have a size of zero

BEIMA.Backend/MongoService/Device.cs
Kenny-Miller: Yeah it does

BEIMA.Backend/DeviceFunctions/UpdateDevice.cs
josephmoore99: Discussed outside of this PR, for now we will only allow users to upload a single photo rather than a set of photos.

roguedeleter:
> Closes #TBDTaskIDorBugID
> 
> TBDAdditionalPRDetails

Please add the story/task number so we can link this PR correctly.

BEIMA.Backend/DeviceFunctions/AddDevice.cs
AshlynAdamson: The mongo connector is also a singleton, can we be consistent and handle singletons the same way?

josephmoore99: I see how you have set up the unit tests to provide the storage provider, but I'm not seeing how when we launch the application on it's own, how it will figure out how to populate that storage provider? I think this could be why the FTs are failing as well.

BEIMA.Backend/Models/Requests.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        public string DeviceTag { get; set; }
+        public string DeviceTypeId { get; set; }
+        public string Manufacturer { get; set; }
+        public string ModelNum { get; set; }
+        public string SerialNum { get; set; }
+        public int YearManufactured { get; set; }
+        public string Notes { get; set; }
+        public Location Location { get; set; }
+        public Dictionary<string, string> Fields { get; set; } = new Dictionary<string, string>();
```
AshlynAdamson, line 22: Are we going to need image and documents objects here as well?

BEIMA.Backend/Models/Requests.cs
```diff
@@ -0,0 +1,52 @@
...
+        public List<string> DeletedFiles { get; set; } = new List<string>();
```
AshlynAdamson, line 39: Do we need AddFIles as well? Do we need an object for an updated image?

BEIMA.Backend.Test/DeviceFunctions/AddDeviceTest.cs
AshlynAdamson: I agree, just like how the mongo connector is a mock object

AshlynAdamson:
> > Closes #TBDTaskIDorBugID
> > TBDAdditionalPRDetails
> 
> Please add the story/task number so we can link this PR correctly.

And a description so reviewers get a context for the changes.

BEIMA.Backend/MongoService/Device.cs
```diff
@@ -116,14 +157,18 @@ private void CheckNullArgument(dynamic arg, string name = "")
...
         {
-            CheckNullArgument(DeviceTag, nameof(DeviceTag));
-            CheckNullArgument(Manufacturer, nameof(Manufacturer));
-            CheckNullArgument(ModelNum, nameof(ModelNum));
-            CheckNullArgument(SerialNum, nameof(SerialNum));
-            CheckNullArgument(YearManufactured, nameof(YearManufactured));
-            CheckNullArgument(Notes, nameof(Notes));
-            CheckNullArgument(Location, nameof(Location));
-            CheckNullArgument(LastModified, nameof(LastModified));
+            if(DeviceTag == null
```
roguedeleter, line 160: Why was the CheckNullArgument() removed? If we just throw a generic ArgumentNullException, then we lose a lot of details in our Exception messages.

BEIMA.Backend/Models/Requests.cs
Kenny-Miller: No as those will exist outside of the data object that is being sent up. In a mulitpart form request, added files and photos will be found in req.Files. The difference is the name property attached to each individual file. Its a little weird, but that is a quirk of the request. The Request.cs objects will map to req.Form['data'] and will be how we are able to send the whole device json alongside the files.

BEIMA.Backend.Test/DeviceFunctions/AddDeviceTest.cs
Kenny-Miller: Sounds good, will get that switched to using the mocker.

BEIMA.Backend/DeviceFunctions/AddDevice.cs
Kenny-Miller: Startup.cs gets ran automatically on startup by the application which is how it gets created. 

The FT's are failing because I'm still in the process of getting the mulitpartform request created correctly. That was why I opened the pullrequest early so I could get a jumpstart on feedback while trying to debug it. The request gets sent fine, but req.ReadFormAsync returns null. Manually hitting the endpoints from postman works just fine, same when running the normal tests so I don't think the dependency injection is the issue.

BEIMA.Backend/MongoService/Device.cs
```diff
@@ -166,31 +211,41 @@ public void SetLocation(ObjectId buildingId, string notes, string latitude, stri
         public void SetLastModified(DateTime? date, string user)
         {
             //Check if null, if they are, then use defualt values
-            date ??= DateTime.Now.ToUniversalTime();
+            var modifiedDate = (date ??= DateTime.Now.ToUniversalTime());
```
roguedeleter, line 214: Why do we need to create a temp variable here?

BEIMA.Backend/DeviceFunctions/AddDevice.cs
josephmoore99: Ah ok, that makes sense. I do still agree with Ashlyn though in that I think it would be in our best interest to be consistent with how we are using the MongoConnector in the endpoints, do you think that it is possible to adjust it so that we don't need to use the constructor? Or perhaps even the other way around?

BEIMA.Backend/MongoService/Device.cs
josephmoore99: As a side note, I think you can also use DateTime.UtcNow instead of DateTime.Now.ToUniversalTime().

BEIMA.Backend/DeviceFunctions/AddDevice.cs
Kenny-Miller: Yeah that would be possible

BEIMA.Backend/MongoService/Device.cs
Kenny-Miller: It didn't like assigning  DateTime? to a DateTime

josephmoore99: If you include the ??= it should be fine no?

josephmoore99: You could try using:
LastModified.Date date = date ?? DateTime.UtcNow;

BEIMA.Backend/MongoService/Device.cs
Kenny-Miller: Added that back in

BEIMA.Backend.FT/DeviceFT.cs
```diff
@@ -92,6 +92,12 @@ public async Task DeviceNotInDatabase_AddDevice_CreatesNewDevice()
             Assert.That(getDevice.Location?.Notes, Is.EqualTo(device.Location.Notes));
             Assert.That(getDevice.Location?.Longitude, Is.EqualTo(device.Location.Longitude));
             Assert.That(getDevice.Location?.Latitude, Is.EqualTo(device.Location.Latitude));
+
+            // Cleanup
+            if(getDevice.Id != null)
```
josephmoore99, line 97: Cleanup is already done in the setup method, so that if you run a test, you can still look at what the state of the database is at the end of the test.

BEIMA.Backend.FT/TestObjects.cs
```diff
@@ -8,7 +8,7 @@ public class TestObjects
     {
         public class Device
         {
-            [JsonProperty(PropertyName = "_id")]
+            [JsonProperty(PropertyName = "id")]
```
josephmoore99, line 11: I think this might cause some of our front end testing to fail, but I could be wrong.

BEIMA.Backend.Test/DeviceFunctions/DeleteDeviceTest.cs
```diff
@@ -17,17 +20,21 @@ public class DeleteDeviceTest : UnitTestBase
...
+
             var request = CreateHttpRequest(RequestMethod.POST);
             var logger = (new LoggerFactory()).CreateLogger("Testing");
 
             // ACT
-            var response = DeleteDevice.Run(request, id, logger);
+            var response = await new DeleteDevice(storageProvider).Run(request, id, logger);
 
             // ASSERT
             Assert.DoesNotThrow(() => mockDb.Verify(mock => mock.DeleteDevice(It.IsAny<ObjectId>()), Times.Never));
```
josephmoore99, line 40: You should also make sure to add an assert here to make sure that the mockStorage didn't get called. (You can look at this mockDb.Verify as an example.)

BEIMA.Backend/MongoService/Device.cs
roguedeleter: You can use this:

date ??= DateTime.UtcNow;

And then when you actually set the Date property:

LastModified.Date = (DateTime)date;

BEIMA.Backend/DeviceFunctions/DeleteDevice.cs
```diff
@@ -35,6 +49,31 @@ public static class DeleteDevice
...
+            var deviceDocument = mongo.GetDevice(deviceId);
+            if(deviceDocument == null)
+            {
+                return new NotFoundObjectResult(Resources.DeviceNotFoundMessage);
+            }
+
+            var device = BsonSerializer.Deserialize<Device>(deviceDocument);
+
+            // Remove Files
+            foreach (var file in device.Files)
```
josephmoore99, line 61: We might need a null check for this.

BEIMA.Backend.FT/TestObjects.cs
Kenny-Miller: Do the frontend tests call the ft tests?

BEIMA.Backend.Test/RequestFactory.cs
Kenny-Miller: Moved those to a method argument

BEIMA.Backend/MongoService/Device.cs
Kenny-Miller: Changed to DateTime.UtcNow. Could have sworn it didn't like the assignment but it works just fine now.

BEIMA.Backend.FT/TestObjects.cs
Kenny-Miller: I'm wrong, I didn't notice that since I changed the device model to serialize back into a c# object it would return id and not _id.

Kenny-Miller:
This branch has been abandoned. It has been split up into smaller pr's instead

PR #234, opened 2022-03-13
Adding maxLength attribute to all Form.Control elements

### LINKED ISSUE #163: Error message that the max character limit has been exceed
#157
We need to check that input fields haven't exceeded 1024 characters and should stop the user from updating/adding fields with that exceeded character amount.

### LINKED ISSUE #238: Create constants file
References user story #237 

add constants file

AshlynAdamson:
Closes #163 
Closes #238 

Technically, there's no error message, but instead I added a maxLength attribute to all Form.Control elements. That way a user can't even insert more than 1024 characters. This obviously easier, but if we want an error message or something similar I can get it added.


roguedeleter:
Can we store this in a constant variable? I don't know if there already exists a constants.js file for the front end, but it would probably be easier to maintain over time to only have this variable in one place.

AshlynAdamson:
I didn't add any tests for this because 1024 characters is a lot of have hard coded in the tests

> 

Good idea, I can get one added

AshlynAdamson:
> Can we store this in a constant variable? I don't know if there already exists a constants.js file for the front end, but it would probably be easier to maintain over time to only have this variable in one place.

Good idea, I can get one created

AshlynAdamson:
> > Can we store this in a constant variable? I don't know if there already exists a constants.js file for the front end, but it would probably be easier to maintain over time to only have this variable in one place.
> 
> Good idea, I can get one created

Constants file now added

roguedeleter (approved):

josephmoore99 (approved):

PR #236, opened 2022-03-13
Create POCO for User

### LINKED ISSUE #113: Create C# object for User
#80 

We need a C# object for a User. This should allow us to convert data into a BsonDocument.

This is slated for Sprint 2.

roguedeleter:
Closes #113 

This PR implements a POCO for User, as well as the relevant unit tests.

### DIFF
--- a/BEIMA.Backend/MongoService/User.cs
+++ b/BEIMA.Backend/MongoService/User.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
 using MongoDB.Bson;
 using MongoDB.Bson.Serialization.Attributes;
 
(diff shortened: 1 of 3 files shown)


BEIMA.Backend/MongoService/User.cs
```diff
@@ -0,0 +1,122 @@
+﻿using System;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+
+namespace BEIMA.Backend.MongoService
+{
+    /// <summary>
+    /// Object representation of a last modified document in a User document
+    /// </summary>
+    public class UserLastModified
```
josephmoore99, line 10: Since this is used by multiple things, I think it would be worth it to put this in its own class file that all the database types can reference.

roguedeleter: Should we do this down the road? I know Kenny made some changes to Device.cs that does the same thing in his PR, so it might be a pain to try and coordinate changing over to a single class file right now. It might be easier to leave it as is right now and change them over all at once after everything is merged in.

josephmoore99 (approved):

thessbsu (approved):

PR #245, opened 2022-03-14
Creating add building page

### LINKED ISSUE #240: Create page for adding a new building
References user story #167 

We need a page to be able to add new buildings to the database

AshlynAdamson:
Closes #240 

Created an add building page, it's pretty simple. Currently, can only access it via the url, but I'm planning to have a button on the building list page that navigates the user to this page. 

image

### DIFF
--- a/BEIMA.Client/src/App.js
+++ b/BEIMA.Client/src/App.js
@@ -15,6 +15,7 @@ import DeviceTypesPage from './pages/DeviceTypes/DeviceTypesPage';
 import DeviceTypePage from './pages/DeviceTypes/DeviceTypePage';
 import AddDevicePage from './pages/Devices/AddDevicePage';
 import AddDeviceTypePage from './pages/DeviceTypes/AddDeviceTypePage';
+import AddBuildingPage from './pages/Building/AddBuildingPage';
 import NavBar from './shared/NavBar';
 import PageTitle from './shared/PageTitle';
 import 'bootstrap/dist/css/bootstrap.min.css';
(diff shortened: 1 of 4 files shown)


BEIMA.Client/src/pages/Building/AddBuildingPage.js
```diff
@@ -0,0 +1,97 @@
...
+import { IoArrowBack } from "react-icons/io5";
+import { Card, Button, Form, Row, Col } from 'react-bootstrap';
+import { useEffect, useState } from "react";
+import styles from './AddBuildingPage.module.css';
+import FormListWithErrorFeedback from '../../shared/FormList/FormListWithErrorFeedback.js';
+
+
+const AddBuildingPage = () => {
+  // this will be replaced with API call based on selected device type to get the fields
+  const mandatoryBuildingFields = {
```
josephmoore99, line 11: I don't think it's necessary to have these referred to as "mandatory" as these should be the same fields for every building object.

BEIMA.Client/src/App.js
```diff
@@ -31,6 +32,7 @@ const App = () => {
           <Route path="devices/:id" element={<DevicePage/>}/>
           <Route path="deviceTypes" element={<DeviceTypesPage/>}/>  
           <Route path="deviceTypes/:typeId" element={<DeviceTypePage/>}/>
+          <Route path="buildings/addBuilding" element={<AddBuildingPage/>}/>
```
josephmoore99, line 35: Should we add a link on the navbar for this page?

AshlynAdamson: I'm going to add one for buildings list, and then from that page there will be an "Add Building" button that will lead to this page

AshlynAdamson: But we can add it to the nav bar, I just don't want to add too much, and the buildings will probably be something that isn't modified a whole lot, just to get the current buildings added.

BEIMA.Client/src/pages/Building/AddBuildingPage.js
AshlynAdamson: Updated the name to allBuildingFields

josephmoore99 (approved):

roguedeleter (approved):

Issue #157, opened 2022-03-03
Add Front End Warning Messages


roguedeleter:
- [x] Prompt the user if they are trying to insert blank fields on Device and DeviceType
- [x] Prompt the user if they are trying to delete fields. Make sure it shows how many devices currently has that field populated. Make them have to confirm they are absolutely sure they understand what they are about to do.
- [x] Max character limit on all fields (1024 proposed)
- [x] Check to see the latitude and longitude is actually in a valid format, throw a hard error if it is not valid.

We proposed to do a sanity check on location. This would involve checking to see if the device is too far away from campus, and prompt the user if they are sure the coordinates are correct. This is currently an optional task.

Acceptance Criteria:

- [x] #159
- [x] #160
- [x] #161
- [x] #162
- [x] #163
- [x] #164
- [x] #333
- [x] #355

PR #268, opened 2022-03-16
Allows device to be added to the DB

### LINKED ISSUE #168: Add Device Page  - Clicking Add Device should add it to the DB

thessbsu:
Closes #168

allows addition of a device to the DB
It still needs building selection integration but that needs to be built out


BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -62,83 +72,151 @@ const AddDevicePage = () => {
...
-      setErrors({});
-      setSelectedDeviceType('Select Device Type');
-      for(let i = 0; i < formFields.length; i++){
-        formFields[i].value = "";
-      }
+      /*TODO Need ability to get a building ID, probably a DD on the page*/
+
+      dbJson.deviceTypeId = selectedDeviceType._id;
       
+      AddDevice(dbJson).then(response => {
```
josephmoore99, line 192: Given that the name of this function is 'createJson', I think we should split this part out into it's own function and just return the JSON object. Then in the add button onClick stuff we can call this chunk of code immediately after the createJson has created the json for us to pass in. Or we could just rename this function to be more descriptive about what it's doing, but I like the idea of splitting up the code a bit so it's not all in one place.

BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -62,83 +72,151 @@ const AddDevicePage = () => {
...
+        if(response.status === HTTP_SUCCESS){
+          setErrors({});
+          setSelectedDeviceType(defaultDeviceTypeObj);
+          setDdStyle(styles.ddSelected);
+          for(let i = 0; i < formFields.length; i++){
+            formFields[i].value = "";
+          }
+        } else {
+          /*TODO push error to display*/
+          alert('API responded with: ' + response.status + ' ' + response.response);
```
josephmoore99, line 203: It is likely that our users won't understand what this means, we should say something like 'Adding the device failed: response details...' instead of 'API responded with'.

thessbsu: The signaling can be improved, I don't like the alert function either it's basic and looks unpolished. Since this coming sprint is "longer" because of the break I was thinking of getting a status/error section inserted into the top of the page. I was thinking a shared component that can be added to any page that needs it with some styling.

Even the response details are sparse from the API (status 400, text Bad Request) so the wording/suggested action will need some massaging. 

Some tasks can be added to #157 for the next sprint to polish it, but at the moment I feel it satisfies the goal of the task to get it working with the API calls to add a device.

We'll need a task to update it for the building part too since it was only built last sprint as well.

BEIMA.Client/src/pages/Devices/AddDevicePage.js
thessbsu: - [ ] split json creation/submit action/change function name TBD

BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -62,83 +72,151 @@ const AddDevicePage = () => {
   const getFieldsForTypeId = async (deviceTypeId) => {
     const deviceTypeFields = await GetDeviceType(deviceTypeId);
     
-    //change the dropdown text
-    setSelectedDeviceType(deviceTypeFields.response.name);
-    
-    let allDeviceFields = mandatoryDeviceFields;
+    //change the dropdown text and store the fields for their keys
+    setSelectedDeviceType(deviceTypeFields.response);
+    setDdStyle(styles.ddSelected);
```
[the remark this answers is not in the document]
thessbsu: - [ ] expand dropdown abbreviation, should prevent it being confused with Bootstrap default "ddselected" (which makes it green) as well vs ddSelected

AshlynAdamson (dismissed):
Besides Joe's comment, I think this looks good!

BEIMA.Client/src/pages/Devices/AddDevicePage.js
thessbsu: createJson function only creates the save data JSON
saveDeviceToDb now calls the createJson function and submits it to the DB

AshlynAdamson (dismissed):

AshlynAdamson (approved):

josephmoore99 (approved):

PR #298, opened 2022-03-17
Force user to enter numeric value for year manufactured 

### LINKED ISSUE #297: Force year manufactured to int on frontend and have proper error message
References user story #82 

If a user enters a string for the year manufactured field, the add device (and I'm sure update device) endpoint will not accept it. We should force this field to an int in all areas where it can be added/changed. 

There should also be an error message for if the user enters non-numeric values and not allow the device to be updated/added.

AshlynAdamson:
Closes #297

In the add device page, if the year manufactured value contains any non-numeric chars, it will show an error message to the user and won't let the user add the device. When editing a device, the user will not be able to input any non-numeric values. 
image

The destination branch is Tom's PR, and there are many diffs because his branch isn't up to date with master. That's okay, we can wait for his PR to merge before we merge this one.

### DIFF
--- a/BEIMA.Client/cypress/integration/AddDevicePage.spec.js
+++ b/BEIMA.Client/cypress/integration/AddDevicePage.spec.js
@@ -24,6 +24,8 @@ describe("Verify Data can be entered into fields", () => {
 describe("Verify Data in fields is cleared when Add Device is selected", () => {
   it('Enter data, click Add Device, verify fields are empty', () => {
     cy.visit('http://localhost:3000/addDevice')
+    cy.get("#typeDropDown").scrollIntoView().click()
+    cy.get("#typeDropDown").children().last().scrollIntoView().click()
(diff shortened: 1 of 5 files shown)


josephmoore99:
I think you need to merge in main?

AshlynAdamson:
> I think you need to merge in main?

Now that Tom's PR is merged, I changed the destination branch to main

thessbsu (dismissed):

AshlynAdamson:
- [ ] only allow 4 chars for year

AshlynAdamson:
Looks like we do not need to force it to an int type on the front end. As long as the value only contains numeric values, the deserialization on the backend can still convert it correctly 
image

AshlynAdamson:
Year Manufactured can only be 4 chars long and an error message is shown if not
image

thessbsu (approved):

josephmoore99 (approved):

PR #316, opened 2022-03-18
Added storage to update endpoint

### LINKED ISSUE #232: Update AddDevice and UpdateDevice to accept a MIME request

Kenny-Miller:
Closes #232 
References #233 

Adds storage capabilities to update device endpoint

### DIFF
--- a/BEIMA.Backend.FT/TestObjects.cs
+++ b/BEIMA.Backend.FT/TestObjects.cs
@@ -47,10 +47,7 @@ namespace BEIMA.Backend.FT
             public List<DeviceFile>? Files { get; set; }
 
             [JsonProperty(PropertyName = "photo")]
-            public DeviceFile? Photo { get; set; }
-
-            [JsonProperty(PropertyName = "deletedFiles")]
-            public List<string>? DeletedFiles { get; set; }
+            public DeviceFile? Photo { get; set; }            
         }
 
         public class DeviceFile
@@ -60,7 +57,7 @@ namespace BEIMA.Backend.FT
             [JsonProperty(PropertyName = "fileUid")]
             public string? FileUid { get; set; }
             [JsonProperty(PropertyName = "fileUrl")]
-            public string? Url { get; set; }
+            public string? FileUrl { get; set; }
         }
 
         public class LastModified
@@ -103,6 +100,11 @@ namespace BEIMA.Backend.FT
             public string? Notes { get; set; }
         }
 
+        public class DeviceUpdate : Device
+        {
+            [JsonProperty(PropertyName = "deletedFiles")]
+            public List<string>? DeletedFiles { get; set; }
+        }
         public class DeviceTypeAdd : DeviceTypeBase
         {
             [JsonProperty(PropertyName = "fields")]
--- a/BEIMA.Backend.FT/TestObjects.cs
+++ b/BEIMA.Backend.FT/TestObjects.cs
@@ -1,11 +1,13 @@
 ﻿using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
+using System.Text;
 
 namespace BEIMA.Backend.FT
 {
     public class TestObjects
     {
+        public static readonly byte[] _fileBytes = Encoding.ASCII.GetBytes("TestOne");
         public class Device
         {
             [JsonProperty(PropertyName = "_id")]
@@ -46,6 +48,9 @@ namespace BEIMA.Backend.FT
 
             [JsonProperty(PropertyName = "photo")]
             public DeviceFile? Photo { get; set; }
+
+            [JsonProperty(PropertyName = "deletedFiles")]
+            public List<string>? DeletedFiles { get; set; }
         }
 
         public class DeviceFile
--- a/BEIMA.Backend.Test/DeviceFunctions/UpdateDeviceTest.cs
+++ b/BEIMA.Backend.Test/DeviceFunctions/UpdateDeviceTest.cs
@@ -1,5 +1,6 @@
 ﻿using BEIMA.Backend.DeviceFunctions;
 using BEIMA.Backend.MongoService;
+using BEIMA.Backend.StorageService;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using MongoDB.Bson;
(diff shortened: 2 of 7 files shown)


BEIMA.Backend.Test/DeviceFunctions/UpdateDeviceTest.cs
```diff
@@ -24,20 +25,29 @@ public async Task IdNotInDatabase_UpdateDevice_ReturnsNotFound()
             // Setup mock database client.
             Mock<IMongoConnector> mockDb = new Mock<IMongoConnector>();
             var testId = "abcdef123456789012345678";
-            mockDb.Setup(mock => mock.UpdateDevice(It.Is<BsonDocument>(bd => bd["_id"].AsObjectId.ToString().Equals(testId))))
```
josephmoore99, line 27: We still need the mock method for UpdateDevice no?

BEIMA.Backend/DeviceFunctions/UpdateDevice.cs
```diff
@@ -40,42 +42,98 @@ public static class UpdateDevice
...
+            device.DeviceTypeId = ObjectId.Parse(data.DeviceTypeId);
+            device.DeviceTag = data.DeviceTag;
+            device.Manufacturer = data.Manufacturer;
+            device.ModelNum = data.ModelNum;
+            device.SerialNum = data.SerialNum;
+            device.YearManufactured = data.YearManufactured;
+            device.Notes = data.Notes;
+
+            var reqBuildingId = data.Location.BuildingId;
+            ObjectId? buildingId = reqBuildingId != null ? ObjectId.Parse(reqBuildingId) : null;
```
josephmoore99, line 83: Does the GetBsonDocument() method throw an exception if this is null?

BEIMA.Backend.Test/DeviceFunctions/UpdateDeviceTest.cs
josephmoore99: Actually, it doesn't even look like we call GetDevice in the update endpoint, I am confused?

josephmoore99: I guess we technically don't actually need it since we are asserting that it is run Times.Never below. I think I'm ok with this actually.

BEIMA.Backend.Test/DeviceFunctions/UpdateDeviceTest.cs
```diff
@@ -52,19 +62,27 @@ public async Task InvalidId_UpdateDevice_ReturnsInvalidId(string id)
...
+
+            var data = TestData._testUpdateDeviceDeleteFiles;
+            var request = CreateMultiPartHttpRequest(data);
             var logger = (new LoggerFactory()).CreateLogger("Testing");
 
             // ACT
             var response = await UpdateDevice.Run(request, id, logger);
 
             // ASSERT
             Assert.DoesNotThrow(() => mockDb.Verify(mock => mock.GetDevice(It.IsAny<ObjectId>()), Times.Never));
```
josephmoore99, line 84: We should assert on Times.Never for mock.UpdateDevice as well.

BEIMA.Backend.Test/DeviceFunctions/UpdateDeviceTest.cs
```diff
@@ -97,6 +125,7 @@ public async Task ExistingDevice_UpdatesDevice_ReturnsOKResponse()
             Assert.IsNotNull(response);
             Assert.That(response, Is.TypeOf(typeof(OkObjectResult)));
             Assert.That(((OkObjectResult)response).StatusCode, Is.EqualTo((int)HttpStatusCode.OK));
+            Assert.DoesNotThrow(() => mockStorage.Verify(mock => mock.DeleteFile(It.IsAny<string>()), Times.Exactly(2)));
```
josephmoore99, line 128: We should have a mock method for DeleteFile if we are going to be asserting on it. Also as a quick note, on the mocks you can have a double returns which will allow you to let it return two values for different inputs, or if you are expecting two different inputs, you can make two separate mock methods to run depending on the input. Sorry if this is a bit confusing.

BEIMA.Backend/DeviceFunctions/UpdateDevice.cs
Kenny-Miller: Yep it works, there is a test called ExistingDevice_UpdatesDeviceWithNoLocation_ReturnsOKResponse that verifies this.

BEIMA.Backend.FT/TestObjects.cs
```diff
@@ -46,6 +48,9 @@ public class Device
 
             [JsonProperty(PropertyName = "photo")]
             public DeviceFile? Photo { get; set; }
+
+            [JsonProperty(PropertyName = "deletedFiles")]
```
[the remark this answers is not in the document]
Kenny-Miller: Moved to DeviceUpdate object

BEIMA.Backend/DeviceFunctions/UpdateDevice.cs
```diff
@@ -40,42 +42,98 @@ public static class UpdateDevice
...
+                    log.LogInformation($"Failed to delete a file: {device.Photo.FileUid.ToString()}.");
+                }
+                device.Photo = null;
+            }
+
+            // Remove files from device's file list
+            device.Files = device.Files.FindAll(file => !data.DeletedFiles.Contains(file.FileUid));
+
+            // Removed deleted files from storage
+            foreach (var fileUid in data.DeletedFiles)
```
josephmoore99, line 111: Should we consider waiting until after the database has been successfully updated to delete these?

Kenny-Miller: That's a really good idea. That way we can ensure the references are removing before deleting the files.

BEIMA.Backend/DeviceFunctions/UpdateDevice.cs
```diff
@@ -40,42 +43,85 @@ public static class UpdateDevice
...
+            
+            // Add new files and photos
+            foreach (var file in reqForm.Files)
+            {
+                var fileUid = await _storage.PutFile(file);
+                if (file.Name == "files")
+                {
+                    device.AddFile(fileUid, file.FileName);
+                }
+                else if (file.Name == "photos")
```
roguedeleter, line 116: "photo" or "photos"?

josephmoore99 (dismissed):

BEIMA.Backend/DeviceFunctions/UpdateDevice.cs
```diff
@@ -40,60 +43,109 @@ public static class UpdateDevice
...
-                    data.Location.Latitude,
-                    data.Location.Longitude
-                );
+            if (!ObjectId.TryParse(data.DeviceTypeId, out _))
+            {
+                return new BadRequestObjectResult(Resources.InvalidIdMessage);
+            }            
+
+            // Set Data to new values
+            device.DeviceTypeId = deviceTypeId;
```
josephmoore99, line 85: What is the reasoning for moving this out here/not using the constructor?

Kenny-Miller: It was mostly I wanted to move all of the parsing objectId logic into the trycatch, but still keeps settings all the values in one spot. That and if the parsing failed I felt that the couldnotparsebody was a fitting error message.

josephmoore99: But this is outside of the trycatch though right?

Kenny-Miller: device type id gets assigned inside the trycatch at line 66

josephmoore99: Yeah, but isn't it safer to include all of these device assignments inside the trycatch? Just so that in the off chance something does actually go wrong here, we still return a could not parse message.

Kenny-Miller: Fair enough. Moved it back inside

josephmoore99 (approved):

roguedeleter (approved):

PR #321, opened 2022-03-19
Story#165 add user management

### LINKED ISSUE #246: add form for adding a new user
References user story ##165

need to have a form to add a user

### LINKED ISSUE #247: add form for updating a user
References user story #165

need to have a form for updating a user

### LINKED ISSUE #248: add form element for deleting a user
References user story #165

need to have a form element to fire the delete user functionality

### LINKED ISSUE #249: add a list of users and their roles
References user story #165

add elements to see all users and their roles

thessbsu:
Closes #246
Closes #247
Closes #248
Closes #249

Basically a clone of devices/buildings in look, feel, functionality.
I had an idea for it to be one self contained page but it wasn't working out and for continuity's sake it was changed to the current form

Had to rebranch since the branch rename didn't stick on local and remote, it was still "tracking" the original branch name so I updated it to keep with the convention


josephmoore99:
Can you add some screenshots?

thessbsu:
image
image
image
image

BEIMA.Client/src/pages/Users/UserPage.js
```diff
@@ -0,0 +1,169 @@
...
+   * Renders a custom form that enables a user
+   * to update a user's fields
+   * @param user: user json
+   * @returns renderable item
+   */
+  const RenderItem = ({user}) => {
+    const [editable, setEditable] = useState(false);
+
+    const [userId] = useState(user.id);
+    const [username, setUserName] = useState(user.username);
```
roguedeleter, line 79: Inconsistent camel case

BEIMA.Client/src/pages/Users/AddUserPage.js
```diff
@@ -0,0 +1,86 @@
+import styles from './AddUserPage.module.css';
+import { useOutletContext, Link } from 'react-router-dom';
+import { IoArrowBack } from "react-icons/io5";
+import { Card, Button, Form, Row, Col } from 'react-bootstrap';
+import FormListWithErrorFeedback from '../../shared/FormList/FormListWithErrorFeedback.js';
+import { useEffect, useState } from "react";
+
+const AddUserPage = () => {
+  const mandatoryUserFields = {
```
AshlynAdamson, line 9: I was going to suggest changing this to allUserFields, but maybe a suggested improvement could be that certain roles might have unique fields, in which this variable name is fine

BEIMA.Client/src/pages/Users/AddUserPage.js
```diff
@@ -0,0 +1,86 @@
...
+        Object.assign(fieldValues, formJSON);
+      }
+    }
+
+    setFullUserJSON(fieldValues);
+    
+    // replace with user api calls when ready
+    console.log(fullUserJSON);
+
+    if ( Object.keys(newErrors).length > 0 ) {
```
AshlynAdamson, line 47: I can't think of any errors they could encounter when filling out this information. Maybe it's not needed on this form?

AshlynAdamson:
Should we allow a user's info to be changed once it's in the DB? I'm wondering if that would add more overhead for authentication stuff. I'm leaning towards not allowing them to change it, but users can still be deleted. For a future task, we should not allow the current user to delete their own entry lol, or maybe only certain roles and delete/maybe update users

thessbsu:
If we go strictly off of a user ID we should be ok allowing everything else to change. If someone uses firstLast as a username and then changes their marital status they might want to change their username. That's the main scenario I can think of for wanting to update that. As long as the username still corresponds to the unique userId... it should be ok?

BEIMA.Client/src/pages/Users/AddUserPage.js
thessbsu: The only thing I could think of is the Role remaining a text field rather than being converted to a dropdown when API integration happens for the page to guard against mistyping a role name. Other than that I agree, and I think changing it to a dropdown is the way to go once we have role admin in place as well.
I think it's safe to remove.

- [ ] remove error check state var/if block

josephmoore99:
> Should we allow a user's info to be changed once it's in the DB? I'm wondering if that would add more overhead for authentication stuff. I'm leaning towards not allowing them to change it, but users can still be deleted. For a future task, we should not allow the current user to delete their own entry lol, or maybe only certain roles and delete/maybe update users

I think there could be the case that a users role could change

BEIMA.Client/src/pages/Users/UserPage.js
thessbsu: good catch, updated

josephmoore99 (dismissed):

AshlynAdamson (dismissed):

roguedeleter (approved):

roguedeleter (dismissed):
I just looked at your screenshot again, and I noticed there is no place to input a password when inserting a new user. I think it will be good to have the field show when inserting, but have it not show when viewing. On edit, show the field, but make sure it is not populated with any data (make it blank).

thessbsu:
image
image
image
image


 - [ ] add password field and validation
 - [ ]  add tests for validation

AshlynAdamson (dismissed):

roguedeleter:
Can you add validation for the username to not be empty as well?

thessbsu:
Should probably force everything. The list will be significantly less informative if there's no first and last since that's what's displayed, or at the very least put the first and last in the details and the username in the card title. Maybe Username and Role in the card title with the first and last in the details?

thessbsu:
updated so all fields are required, the really all need to be filled out

josephmoore99 (approved):

AshlynAdamson (approved):

PR #322, opened 2022-03-19
device type update updates devices

### LINKED ISSUE #190: Given a device type update, when updating fields, then the fields on all existing devices with that device type should be updated.
#175

josephmoore99:
Closes #190


BEIMA.Backend/DeviceTypeFunctions/DeleteDeviceType.cs
```diff
@@ -36,6 +37,13 @@ public static class DeleteDeviceType
             var mongo = MongoDefinition.MongoInstance;
             var deviceTypeId = new ObjectId(id);
 
+            // Do not delete if at least one device exists with this device type.
+            // TODO: Use database filter for devices instead of getting the list of all devices.
+            if (mongo.GetAllDevices().Where(d => d["deviceTypeId"].AsObjectId.Equals(deviceTypeId)).Any())
```
josephmoore99, line 42: Decided to also add a check in the delete device endpoint to make sure that we don't delete a device type if a device with that device type exists.

AshlynAdamson: I like having this, just in case the delete button gets selected accidentally. But if the user really wants to delete a device type that has entries, should we still let them to do it? I'm imagining the frontend would prompt them if that's what they want to do, and if they do then allow it

josephmoore99: I don't feel as comfortable deleting devices on device type removals as I am with just removing specific fields on device type updates. I think that if they really want to delete the device type, they should be forced to go in and delete all the associated devices first, but I am open to more discussion.

AshlynAdamson: I like the idea of forcing the user to delete devices first. Just thought I'd bring it up, I am also open to more discussion

AshlynAdamson (approved):

roguedeleter (dismissed):

roguedeleter (approved):

AshlynAdamson (approved):

PR #323, opened 2022-03-19
updated initial dev setup with MongoDB install instructions

### LINKED ISSUE #311: Update initial setup guide with local mogodb setup instructions
References user story #147

Since we've updated to using local mongo instances for development the dev setup guide needs to include steps for installing

thessbsu:
Closes #311

I added instructions for completing the mongo setup in MongoCompass.
I'm not sure if I captured them all since it was a while ago (and via docker on a separate machine) that I set it up so correct/add as needed!

### DIFF
--- a/docs/Initial Setup/InitialSetup.md
+++ b/docs/Initial Setup/InitialSetup.md
@@ -23,6 +23,9 @@ Launch Visual Studio
 * API is ready to accept requests
 
 Launch Mongo Compass
+ * Click New Connection
+ * Type or paste " mongodb://localhost:27017 " into the connection string textbox
+ * Click Connect
  * Click the + sign at the bottom of the left pane to add a database
  * Database name: beima
  * Collection name: deviceTypes
--- a/docs/Initial Setup/InitialSetup.md
+++ b/docs/Initial Setup/InitialSetup.md
@@ -1,8 +1,10 @@
+
 ## Initial setup
 
 Install a git client for Windows
 Install Node JS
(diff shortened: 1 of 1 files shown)


docs/Initial Setup/InitialSetup.md
```diff
@@ -19,3 +21,14 @@ Launch Visual Studio
...
 * Run the project in Debug Any CPU
 * API is ready to accept requests
+
+Launch Mongo Compass
+ * Click New Connection
+ * Type or paste " mongodb://localhost:27017 " into the connection string textbox
+ * Click Connect
+ * Click the + sign at the bottom of the left pane to add a database
+ * Database name: beima
+ * Collection name: deviceTypes
```
josephmoore99, line 31: Should we add devices to this as well?

thessbsu: The initial database creation requires a collection name so I picked the one at the top of the list.
The following steps:

"This will create the DB with one of the required collections.
Once that is completed the + sign inline with the DB name to add the remaining required collections to mirror the production environment."

are there to direct creation of other collections to match the prod environment.

josephmoore99 (approved):

docs/Initial Setup/InitialSetup.md
AshlynAdamson: Also, the backend should create the collections for you, as long as they are specified in the local.settings.json.

AshlynAdamson (approved):

PR #325, opened 2022-03-20
Building POST endpoint

### LINKED ISSUE #286: Given a building, when the building POST endpoint is called, then the new building should be added to the database.
Story #271
Our backend should have an endpoint that allows us to add new buildings to the database.

josephmoore99:
Closes #286

### DIFF
--- a/BEIMA.Backend.FT/BuildingFT.cs
+++ b/BEIMA.Backend.FT/BuildingFT.cs
@@ -19,8 +19,8 @@ namespace BEIMA.Backend.FT
                 Notes = "Some notes.",
                 Location = new Location
                 {
-                    Longitude = "123.001",
-                    Latitude = "101.321",
+                    Longitude = "12.001",
+                    Latitude = "24.321",
                 },
             };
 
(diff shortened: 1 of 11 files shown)


BEIMA.Backend.FT/BuildingFT.cs
```diff
@@ -0,0 +1,36 @@
...
+            // ARRANGE
+            var building = new Building
+            {
+                Name = "Student Union",
+                Number = "1234",
+                Notes = "Some notes.",
+                Location = new Location
+                {
+                    Longitude = "123.001",
+                    Latitude = "101.321",
```
AshlynAdamson, line 23: This lat value out of range, do we need checking for that on the backend, or is the front end validation enough?

josephmoore99: Haven't added the checking for the backend on this yet, I think we are going to do that on another task, but I will update this to values that are actually in range.

AshlynAdamson (approved):

roguedeleter (approved):

PR #328, opened 2022-03-20
Refactor MongoConnector to abstract out basic CRUD calls

### LINKED ISSUE #327: Refactor MongoConnector to abstract CRUD calls
#273

This task is to refactor the MongoConnector to abstract out the CRUD calls. All of these calls can be generalized, and it doesn't make sense to have separate implementations for each database object.

roguedeleter:
Closes #327 

This PR abstracts out a lot of the basic CRUD calls in the MongoConnector. I realized that there does not need to be a separate implementation for each POCO, and it can all be abstracted. This will make adding on new MongoConnector functionality much easier in the future.


josephmoore99 (approved):
Nice, this cuts down on a lot of repeated code.

AshlynAdamson (approved):

Issue #336, opened 2022-03-21
Authentication: JWT Auth Flow


Kenny-Miller:
References user story #16

Create an authentication flow detailing the high level overview of how jwt's will be used in the application.

PR #341, opened 2022-03-22
16 authentication service

### LINKED ISSUE #16: Authentication Service
#2

Since we will make use of JWTs to handle endpoint security and role based frontend permissions, we will need a service that can create and validate JWT tokens. The authentication service should make use of dependency injection to make it into a service each function can use.

[ ] As a developer, I should have a service that can produce a JWT with a name and role.
[ ] As a developer, I should have a service that can validate a JWT against a stored private key.

### LINKED ISSUE #342: Authentication Service Implementation
References user story #16

An authentication service needs to be created so that JWT tokens can be create and parsed on the backend. This service should take in a db user object and create a set of claims that are then encoded into a JWT token. The service should also read in HttpRequests and try to parse a JWT and its claims from the request's headers.

Kenny-Miller:
Closes #342

Implements authentication service that allows a user to be converted into a jwt token as well as the ability to parse the claims of that jwt token.


BEIMA.Backend.Test/AuthenticationService/AuthenticationServiceTest.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using static BEIMA.Backend.Test.RequestFactory;
+
+namespace BEIMA.Backend.Test.AuthService
+{
+    [TestFixture]
+    public class AuthServiceTest : UnitTestBase
```
josephmoore99, line 20: I think the file name should match the class name.

BEIMA.Backend/AuthService/AuthenticationService.cs
```diff
@@ -0,0 +1,88 @@
...
+        public static AuthenticationService Instance { get { return instance.Value; } }
+
+        public string CreateToken(User user)
+        {
+            var claims = new Claims()
+            {
+                Username = user.Username,
+                Role = user.Role,
+            };
+            var secretKey = Environment.GetEnvironmentVariable("JwtKey");
```
josephmoore99, line 42: So, we will be doing this differently when we get logins working right? Doesn't feel super secure to be holding our secret key in an environment variable.

BEIMA.Backend/AuthService/IAuthenticationService.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BEIMA.Backend.AuthService
+{
+    public interface IAuthenticationService
+    {
+        string CreateToken(User user);
```
josephmoore99, line 14: Interface should still have the class/method headers. You can just copy/paste the ones from AuthenticationService.

BEIMA.Backend/Startup.cs
```diff
@@ -1,27 +0,0 @@
-﻿using BEIMA.Backend.StorageService;
-using Microsoft.Azure.Functions.Extensions.DependencyInjection;
-using Microsoft.Extensions.DependencyInjection;
-using System;
-
-[assembly: FunctionsStartup(typeof(BEIMA.Backend.Startup))]
-
-namespace BEIMA.Backend
-{
-    public class Startup : FunctionsStartup
```
josephmoore99, line 10: Why don't we need this anymore?

BEIMA.Backend.Test/AuthenticationService/AuthenticationServiceTest.cs
```diff
@@ -0,0 +1,157 @@
...
+            var authSerivce4 = AuthenticationService.Instance;
+            var authSerivce5 = AuthenticationService.Instance;
+            Assert.That(authSerivce1, Is.EqualTo(authSerivce2));
+            Assert.That(authSerivce1, Is.EqualTo(authSerivce3));
+            Assert.That(authSerivce1, Is.EqualTo(authSerivce4));
+            Assert.That(authSerivce1, Is.EqualTo(authSerivce5));
+        }
+
+        [Test]
+        public void HttpRequestNoHeaders_NoClaimsReturned()
```
josephmoore99, line 38: You can name this something like HttpRequestNoHeaders_Parse_NoClaimsReturned()

BEIMA.Backend.Test/AuthenticationService/AuthenticationServiceTest.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            var authService = AuthenticationService.Instance;
+            User user = new User()
+            {
+                Username = "username",
+                Role = "role"
+            };
+
+            var token = authService.CreateToken(user);            
+            var request = CreateHttpRequest(RequestMethod.GET, authToken:token);
```
josephmoore99, line 68: Wait, where did authToken come from? I'm not too familiar with this kind of stuff lol

BEIMA.Backend.Test/AuthenticationService/AuthenticationServiceTest.cs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                Username = "username",
+                Role = "role"
+            };
+
+            var token = authService.CreateToken(user);
+            var requestOne = CreateHttpRequest(RequestMethod.GET, authToken: token);
+
+            var claims = authService.ParseToken(requestOne);
+            claims.Role = "admin";
```
josephmoore99, line 113: I think it would be appropriate to use some Assume.That()'s here on the claims variable.

BEIMA.Backend.Test/AuthenticationService/AuthenticationServiceTest.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            var authService = AuthenticationService.Instance;
+            User user = new User()
+            {
+                Username = "username",
+                Role = "role"
+            };
+
+            var token = authService.CreateToken(user);
+            var requestOne = CreateHttpRequest(RequestMethod.GET, authToken: token);
```
josephmoore99, line 110: Should be consistent with spacing on the authToken parameter.

josephmoore99:
I think we should be linking PRs to a task instead of a story. Was this meant to link to #336?

BEIMA.Backend/AuthService/AuthenticationService.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (string.IsNullOrEmpty(authHeader))
+            {
+                return null;
+            }
+
+            try
+            {
+                if (authHeader.StartsWith("Bearer"))
+                {
+                    authHeader = authHeader.Substring(7);
```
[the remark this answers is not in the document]
Kenny-Miller: The normal format use is "Bearer token"  and "token"

BEIMA.Backend/Startup.cs
Kenny-Miller: This was a remnant of how I had the storage service setup to use dependency injection. Thought I removed it previously, but I guess not.

BEIMA.Backend/AuthService/AuthenticationService.cs
Kenny-Miller: I think this will have to be a bigger discussion as we will want to move the db/storage connection strings out as well. there is a secrets manger that we can use to store theses variables. I'm not sure how it works with github build/deploy though

BEIMA.Backend.Test/AuthenticationService/AuthenticationServiceTest.cs
```diff
@@ -0,0 +1,157 @@
...
+    {
+        [Test]
+        public void ServiceNotCreated_CallConstructorFiveTimes_FiveInstancesAreEqual()
+        {
+            //Tests to make sure singleton is working
+            var authSerivce1 = AuthService.AuthenticationService.Instance;
+            var authSerivce2 = AuthService.AuthenticationService.Instance;
+            var authSerivce3 = AuthService.AuthenticationService.Instance;
+            var authSerivce4 = AuthService.AuthenticationService.Instance;
+            var authSerivce5 = AuthService.AuthenticationService.Instance;
```
Kenny-Miller, line 30: AuthService.AuthenticationService is used because of namespace conflicts between Test.AuthenticationService
and AuthService.AuthenticationService. This only needed in the tests and the 'AuthService.' will not appear in the normal functions

Kenny-Miller:
> I think we should be linking PRs to a task instead of a story. Was this meant to link to #336?

Yeah I miss linked it. Should be linked to the correct task now.

BEIMA.Backend.Test/AuthenticationService/AuthenticationServiceTest.cs
josephmoore99: I think the issue is that your folder/namespace name is the same as your class name. If you look at the MongoConnector/MongoConnectorTest, they are in folders/namespaces called MongoService. I would suggest possibly using Authentication or AuthService as the folder names, and AuthenticationService/AuthenticationServiceTest as the class names, and then I think your namespace conflicts will be resolved.

BEIMA.Backend.Test/AuthenticationService/AuthenticationServiceTest.cs
josephmoore99: - [ ] You missed one method

josephmoore99 (approved):

BEIMA.Backend/AuthService/AuthenticationService.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                return null;
+            }
+
+            try
+            {
+                // Strip of frontal 'Bearer ' if it exists
+                if (authHeader.StartsWith("Bearer "))
+                {
+                    authHeader = authHeader.Substring(7);
```
roguedeleter, line 80: Is this really the best way to do this? 7 seems like a magic number

roguedeleter (approved):

BEIMA.Backend/AuthService/AuthenticationService.cs
Kenny-Miller: Its the length of "Bearer ". The normal protocol is either "token" or "Bearer token", so this just strips off the "Bearer ".

PR #344, opened 2022-03-23
Task 315 - shared components for notifications

### LINKED ISSUE #315: shared component for success/warning/error messages
References user story #157

need to add a shared component that can be dropped into pages that takes in a message type, text, and a flag for hard or soft stop
warnings will be soft stops "user can proceed, but they must confirm it's what they want to do" with a button to confirm
hard stops will be "unable to proceed until fixed" and will have no path to continue

thessbsu:
Closes #315

Added SweetAlert2 package and simplified the 3 main calls we'll use (success, warning, error) through the Notifications.js file to take care of the common parameters without having to from the whole message each time

Screenshot 2022-03-22 185455
Screenshot 2022-03-22 185527
Screenshot 2022-03-22 185555
Screenshot 2022-03-22 185632

### DIFF
--- a/BEIMA.Client/package-lock.json
+++ b/BEIMA.Client/package-lock.json
@@ -24,6 +24,9 @@
         "react-router-dom": "^6.2.1",
         "react-scripts": "^5.0.0",
         "start-server-and-test": "^1.14.0",
+        "sweetalert2": "^11.4.6",
+        "sweetalert2-react-content": "^4.2.0",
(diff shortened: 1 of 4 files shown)


josephmoore99 (approved):
Nice, these will make it much more clear to users what is going on. I like that you added the checkmark for when the device is successfully added as well.

AshlynAdamson (approved):

PR #347, opened 2022-03-24
Photo/files frontend integration

### LINKED ISSUE #280: Update Add Device page to send a photo/documents

### LINKED ISSUE #281: Update View Device page to update photo/documents

### LINKED ISSUE #282: Update View Device page to delete photo/documents

AshlynAdamson:
Closes #280 
Closes #281 
Closes #282 

We can now add a device with an image and additional documents, as well as view those and update/delete them.

Please enjoy this demo
https://user-images.githubusercontent.com/58532471/159938663-96a7ae08-c106-42b2-88aa-0508a33153df.mp4


thessbsu (approved):

run.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
josephmoore99, line 1: Perhaps we should be more descriptive with the name of this script? Maybe something like runBackgroundProcesses.sh, or something along those lines?

BEIMA.Client/src/services/AddDevice.js
```diff
@@ -7,9 +7,21 @@ const API_URL = process.env.REACT_APP_API_URL;
...
+  // setup the multiform request data
   let formData = new FormData();
   formData.append("data", JSON.stringify(deviceDetails));
+  formData.append("photo", photo);
+  const tempFiles = [...files]
+
+  // add each additional file to the files key
+  if(files){
+    tempFiles.map((file, index) => {
+      formData.append(`files${index}`, file);
```
josephmoore99, line 20: Just trying to understand here, so each file is going in their own bucket? I thought we were putting all the extra files in the 'files' bucket? Also, does this include the file names when it sends these over?

run.sh
```diff
@@ -0,0 +1,20 @@
...
+
```
josephmoore99, line 2: It doesn't look like we are calling this script from test.sh? I assume you were meaning to do that?

josephmoore99 (changes requested):
This looks like it's working great looking at the demo! Nice to have the last of the device features pretty much done!

BEIMA.Client/src/pages/Devices/DevicePage.js
```diff
@@ -15,7 +15,7 @@ const DevicePage = () => {
   const [loading, setLoading] = useState(true);
   const [device, setDevice] = useState(null);
   const [image, setImage] = useState(null)
-  const [documents, setDocuments] = useState([])
+  const [documents, setDocuments] = useState({})
```
Kenny-Miller, line 18: This should still be an array as when it gets set by setDocuments(device.files), it'll become an array again as device.files is an array of json objects.

BEIMA.Client/src/pages/Devices/DevicePage.js
```diff
@@ -171,11 +170,13 @@ const DevicePage = () => {
     const [long, setLong] = useState(device.location.longitude)
     const [locNotes, setLocNotes] = useState(device.location.notes)
 
-    const [imageCopy, setImageCopy] = useState(image)
+    const [newImage, setNewImage] = useState()
+    const [imageCopyUrl, setImageCopyUrl] = useState(image.fileUrl)
```
Kenny-Miller, line 174: I would change this from imagecopy to something like previewimage... just because its set by both the device image as well as the upload photo input.

run.sh
Kenny-Miller: I think this is mostly just to spin everything up locally in one script just for local development

Kenny-Miller:
image
These will need to be set to null and [], otherwise the add call will throw an exception if no images/files are added.

BEIMA.Client/src/pages/Devices/DevicePage.js
```diff
@@ -275,36 +268,50 @@ const DevicePage = () => {
...
+        setNewImage(event.target.files[0])
       }
     }
 
+    const deleteImage = () => {
+      let delDocs = deletedDocs;
+      delDocs.push(imageCopy.fileUid);
+
+      setDeletedDocs(delDocs);
+      setNewImage(new File([""], ""))
```
Kenny-Miller, line 281: I would instead set the new image to null, and add the photo's file uid to deletedDocs, or make it its own key in the data object in the update request. Then on the backend depending on the implementation check if we need to delete the photo first then set the new photo if any.

BEIMA.Client/src/services/AddDevice.js
Kenny-Miller, line 20: I was wrong with this piece, doing formData.append(`files`, file); actually does what we want, and it means we can get rid of the contains("files") change on the backend. Can someone sanity check this though, as I thought I tested doing formData.append(`files`, file); and it didn't work, but doing it now does.

run.sh
AshlynAdamson: Yes, this is more for convenience since we need three things running now for BEIMA to work

BEIMA.Client/src/services/AddDevice.js
AshlynAdamson: Yes I believe that is still correct, the files going into one bucket. I started off using the code Kenny suggested in the chat, which had it like "files[${index}]", but it looked like the backend read each file with the name in the formData as "files[0]" etc. I can try removing the index variable and just appending to the files, I don't think append overrides anything in there

- [ ] try without index

BEIMA.Client/src/pages/Devices/DevicePage.js
AshlynAdamson: Good catch, thanks! I'm not sure how it is working now, but I'll play around with it

BEIMA.Client/src/pages/Devices/DevicePage.js
AshlynAdamson: Good idea

- [ ] change name of image url variable to something more clear

AshlynAdamson:
> image
> These will need to be set to null and [], otherwise the add call will throw an exception if no images/files are added.

- [ ] do this

BEIMA.Client/src/pages/Devices/DevicePage.js
AshlynAdamson: I will set this to NULL instead. In the deleteImage function, I do add the uid to the deletedFiles variable and that does work.

- [ ] Change the new image object to null if there is no image

BEIMA.Client/src/services/AddDevice.js
AshlynAdamson: Sweet, I made a similar comment to one of Joe's comments about this. I will give it a try

run.sh
AshlynAdamson: Renamed to runBEIMAProcesses.sh since it runs everything needed for BEIMA

BEIMA.Client/src/services/AddDevice.js
AshlynAdamson: Appending to files works

BEIMA.Backend/DeviceFunctions/AddDevice.cs
```diff
@@ -99,7 +99,7 @@ public static class AddDevice
                     var fileUid = await _storage.PutFile(file);
                     device.SetPhoto(fileUid, file.FileName);
                 }
-                else if (file.Name == "files" && file.Length > 0)
+                else if (file.Name.Contains("files") && file.Length > 0)
```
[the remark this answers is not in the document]
AshlynAdamson: I changed it back to how it was before since we are able to append the the files field in the form data

BEIMA.Client/src/services/AddDevice.js
AshlynAdamson: Appending to files seems to work

BEIMA.Client/src/pages/Devices/DevicePage.js
AshlynAdamson: Updated to null

BEIMA.Client/src/pages/Devices/DevicePage.js
AshlynAdamson: Changed variable name to imagePreviewUrl

BEIMA.Client/src/pages/Devices/DevicePage.js
AshlynAdamson: Changed back to array

josephmoore99 (approved):

thessbsu (approved):

Kenny-Miller (approved):

PR #349, opened 2022-03-26
Implement filtering for Device in MongoConnector

### LINKED ISSUE #337: Implement filtering for Device
#273

roguedeleter:
Closes #337 

This PR implements filtering functionality inside the MongoConnector. GetFilteredDevices was also implemented. 

It also abstracts out the Get method to use the GetFiltered method, since that is basically the same logic. The GetEqualsFilter method was also implemented to make retrieving a filter easier. All it does is return an "equals" filter given a key and value.

### DIFF
--- a/BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
+++ b/BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
@@ -193,6 +193,14 @@ namespace BEIMA.Backend.Test.MongoService
         public void CreateFilterWithNoResults_GetFilteredDevices_NoResultsInList(string key, dynamic value)
         {
             var mongo = MongoConnector.Instance;
+            var doc = new BsonDocument
+            {
+                { "item", "isNotInDb" }
+            };
+            //Insert device
+            var insertResult = mongo.InsertDevice(doc);
+            Assume.That(insertResult, Is.Not.Null);
+            Assume.That(insertResult, Is.TypeOf(typeof(ObjectId)));
 
             //Create filter
             var filter = mongo.GetEqualsFilter(key, value);
--- a/BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
+++ b/BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
@@ -163,6 +163,47 @@ namespace BEIMA.Backend.Test.MongoService
             Assert.IsNull(result);
         }
 
+        [TestCase("serialNumber", "a12345")]
+        [TestCase("key1", 12345)]
+        [TestCase("key2", true)]
+        public void InsertDevice_GetFilteredDevices_DeviceInList(string key, dynamic value)
+        {
+            var mongo = MongoConnector.Instance;
+            var doc = new BsonDocument
+            {
+                { key, value }
+            };
+            //Insert device
+            var insertResult = mongo.InsertDevice(doc);
+            Assume.That(insertResult, Is.Not.Null);
+            Assume.That(insertResult, Is.TypeOf(typeof(ObjectId)));
+
+            //GetFiltered
+            var filter = mongo.GetEqualsFilter(key, value);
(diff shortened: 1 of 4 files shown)


BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -104,15 +104,26 @@ private void CheckIsConnected()
...
         private BsonDocument Get(ObjectId objectId, string dbName, string collectionName)
         {
-            CheckIsConnected();
-
             var filter = Builders<BsonDocument>.Filter.Eq("_id", objectId);
+            return GetFiltered(filter, dbName, collectionName)?.FirstOrDefault();
+        }
+
+        /// <summary>
+        /// Gets a single BsonDocument from the given database/collection, given an ObjectId.
```
josephmoore99, line 112: I think we should probably indicate that this can return any number of documents.

BEIMA.Backend/MongoService/MongoConnector.cs
```diff
@@ -417,5 +438,10 @@ public BsonDocument UpdateUser(BsonDocument doc)
         }
 
         #endregion
+
+        public FilterDefinition<BsonDocument> GetEqualsFilter(string key, dynamic value)
```
josephmoore99, line 442: docs

josephmoore99, line 442: Would this make more sense in a separate static filter generator class or something?

BEIMA.Backend.Test/MongoService/MongoConnectorTest.cs
```diff
@@ -163,6 +163,47 @@ public void ConnectorCreated_UpdateNullDevice_NullReturned()
...
+            }
+        }
+
+        [TestCase("InvalidKey1", "aaaaaaaaaaaa")]
+        [TestCase("aaaaaaaaaaaaaa", 12345)]
+        [TestCase("bbbbbbb", true)]
+        public void CreateFilterWithNoResults_GetFilteredDevices_NoResultsInList(string key, dynamic value)
+        {
+            var mongo = MongoConnector.Instance;
+
```
josephmoore99, line 196: I think we should add in a dummy device here just to make sure we aren't retrieving it.

BEIMA.Backend/MongoService/MongoConnector.cs
Kenny-Miller: Yeah this could probably be static.

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: fixed

BEIMA.Backend/MongoService/MongoConnector.cs
roguedeleter: fixed

roguedeleter: I implemented the MongoFilterGenerator static class and moved this method in there.

roguedeleter:

josephmoore99 (approved):

josephmoore99:
Does this filter work for nested values like Device.Location.Latitude etc.?

thessbsu (approved):

roguedeleter:
> Does this filter work for nested values like Device.Location.Latitude etc.?

As of now this does not work. It only supports top level properties. We could implement more functionality in the future however.

roguedeleter:
> > Does this filter work for nested values like Device.Location.Latitude etc.?
> 
> As of now this does not work. It only supports top level properties. We could implement more functionality in the future however.

@josephmoore99 correction, this actually does work. Here's a test I performed showing that it works:
image


Notice that in the GetEqualsFilter call, I use "item1.item2" as the key. It looks like the method automatically parses the periods to indicate that it is a property of another object.

PR #352, opened 2022-03-29
Update building endpoint

### LINKED ISSUE #289: Given a building, when the building update endpoint is called, then the building should be updated in the database.
Story #271
Our backend should have an endpoint that allows us to update buildings in the database.

josephmoore99:
Closes #289

Implements the update building endpoint with associated tests.

### DIFF
--- a/BEIMA.Backend.FT/BuildingFT.cs
+++ b/BEIMA.Backend.FT/BuildingFT.cs
@@ -212,8 +212,8 @@ namespace BEIMA.Backend.FT
                 Notes = "Some building notes.",
                 Location = new Location
                 {
-                    Latitude = "123.456",
-                    Longitude = "101.101",
+                    Latitude = "12.345",
+                    Longitude = "10.101",
                 },
             };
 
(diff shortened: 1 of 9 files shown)


BEIMA.Backend.FT/BuildingFT.cs
```diff
@@ -200,5 +200,57 @@ public async Task BuildingInDatabase_DeleteBuilding_BuildingDeletedSuccessfully(
...
+        {
+            // ARRANGE
+            var origBuilding = new Building
+            {
+                Name = "Student Union",
+                Number = "1234",
+                Notes = "Some building notes.",
+                Location = new Location
+                {
+                    Latitude = "123.456",
```
AshlynAdamson, line 215: The lat values are out of range, we should probably make those valid. Can't remember if we are planning to add lat/long validation to the backend, but would be good to have valid test values if we do

josephmoore99: Forgot about this again, yeah we should have lat/lon validation in by the end of the sprint

AshlynAdamson (approved):

PR #362, opened 2022-03-30
Add combine filter functionality in MongoFilterGenerator

### LINKED ISSUE #360: Add ability to combine filters in MongoFilterGenerator
References user story #273 

This task will involve implementing a new method to combine filters together in the MongoFilterGenerator. This is to allow us to filter using multiple properties, instead of just one. For example, I want prop1 == 0 AND prop2 == true. This cannot be done with current functionality.

roguedeleter:
Closes #360

This PR adds the ability to combine two separate filters into one filter. Essentially, it performs an AND operation on two filters to combine them together. Now, we will be able to match on multiple properties. 

For example:

username: "user1"
AND
password: "verysecurepassword!"

To use the CombineFilters method, you will need to first create two separate filters using the GetEqualsFilter method. Then, pass the two filters in to the CombineFilters method.

### DIFF
--- a/BEIMA.Backend/MongoService/MongoFilterGenerator.cs
+++ b/BEIMA.Backend/MongoService/MongoFilterGenerator.cs
@@ -27,5 +27,16 @@ namespace BEIMA.Backend.MongoService
             }
             return Builders<BsonDocument>.Filter.Eq(key, value);
         }
+
+        /// <summary>
+        /// Combines two passed in filters. Will return one filter with the AND operation applied between the two separate filters.
+        /// </summary>
+        /// <param name="f1">The first filter.</param>
+        /// <param name="f2">The second filter.</param>
+        /// <returns>One filter that has the AND operation applied to both filters.</returns>
+        public static FilterDefinition<BsonDocument> CombineFilters(FilterDefinition<BsonDocument> f1, FilterDefinition<BsonDocument> f2)
+        {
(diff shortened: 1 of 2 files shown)


BEIMA.Backend/MongoService/MongoFilterGenerator.cs
```diff
@@ -27,5 +27,16 @@ public static FilterDefinition<BsonDocument> GetEqualsFilter(string key, dynamic
...
             return Builders<BsonDocument>.Filter.Eq(key, value);
         }
+
+        /// <summary>
+        /// Combines two passed in filters. Will return one filter with the AND operation applied between the two separate filters.
+        /// </summary>
+        /// <param name="f1">The first filter.</param>
+        /// <param name="f2">The second filter.</param>
+        /// <returns>One filter that has the AND operation applied to both filters.</returns>
+        public static FilterDefinition<BsonDocument> CombineFilters(FilterDefinition<BsonDocument> f1, FilterDefinition<BsonDocument> f2)
```
josephmoore99, line 37: Instead of just limiting it to two at a time, why not add a params argument for the filter definitions here. That way we can use any number:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/params

roguedeleter:

josephmoore99 (approved):

Kenny-Miller (approved):

PR #366, opened 2022-03-31
Task#269 device type list test fails

### LINKED ISSUE #269: Device Types List page test always fails
References user story #82

the css class selector is dynamically generated

**Steps to Reproduce:**
1. run cypress/front end tests on github actions or locally

**Actual Results:**
when the test for cards existing based on css class executes it fails because of the "__XXXX" identifier is added to the class name at runtime

**Expected Results:**
the test should identify the cards existence and pass

**Other notes:**
may need to find another way to select the cards or see if there's a wildcard like operator in cypress for css selection

thessbsu:
Closes #269
based test on classes that have a name that starts with "ItemList_item" and count the results
when at least one device exists in the DB there will be (minimum) 2 divs with the specified CSS identifier

### DIFF
--- a/BEIMA.Client/cypress/integration/DeviceTypesPage.spec.js
+++ b/BEIMA.Client/cypress/integration/DeviceTypesPage.spec.js
@@ -12,9 +12,9 @@ describe("Verify the list has items present", function () {
   it('Count children of item list (see comments)', function (){
     cy.visit('http://localhost:3000/deviceTypes')
 
-    //get all children of the itemlist with class ItemList_item__2BOfJ
-    //there should be at least 1 item in the list
-    //cy.get('[id=itemList]').find('.^="ItemList_item"').its('length').should('be.gt', 0);
+    //get all elements on the page with a class that starts with "ItemList_item"
+    //if there's at least one device in the DB it will pass since 2 elements with
+    //the expected class are created for each device listing
     cy.get('[class^="ItemList_item"]').its('length').should('be.gt', 2);
   })
 })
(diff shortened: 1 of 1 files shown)


AshlynAdamson (approved):
Can we get a screenshot of passing results?

BEIMA.Client/cypress/integration/DeviceTypesPage.spec.js
```diff
@@ -12,9 +12,9 @@ describe("Verify the list has items present", function () {
...
     cy.visit('http://localhost:3000/deviceTypes')
 
-    //get all children of the itemlist with class ItemList_item__2BOfJ
-    //there should be at least 1 item in the list
-    cy.get('[id=itemList]').find('.ItemList_item__2BOfJ').its('length').should('be.gt', 0);
-
+    //get all elements on the page with a class that starts with "ItemList_item"
+    //if there's at least one device in the DB it will pass since 2 elements with
+    //the expected class are created for each device listing
+    cy.get('[class^="ItemList_item"]').its('length').should('be.gt', 2);
```
josephmoore99, line 18: Is this meant to be greater than or greater than/equal to 2.

thessbsu: With 1 device type in the DB there's a minimum of 3 items that will match the selector:
image

So I set the test to look for more than min-1 so if there's 1 device the count will be greater than 2. As I was verifying I think it would be more understandable at >= 3 so it looks for the absolute minimum instead

- [ ] change test to >=3 rather than >2

josephmoore99: I was more going off of the comment, which seemed to imply that it was >= 2, but I could just be misunderstanding

thessbsu:
image

josephmoore99 (approved):

Kenny-Miller (approved):

Issue #275, opened 2022-03-17
Report Service


josephmoore99:
As an admin, I need to be able to generate csv's/reports based on device/device type data, so that I can have insight into the data.

**Acceptance Criteria:**
- [x] Given devices in the database, when the generate export all data button is clicked, then a zip file containing csvs of all device types and their respective devices is downloaded.
- [x] Given filtering devices by a device type, when a download device type csv button is clicked, then a file containing device data for the specific device type is downloaded.

PR #371, opened 2022-03-31
348 report service

### LINKED ISSUE #348: Report Service Provider
References user story #275

This task is to implement a csv service that can be used by endpoints to generate a file/zip containing an export of all devices per device type. This service will initially be able to create one file given a device type, or a zip file containing a file per device type.

Kenny-Miller:
References #275
Closes #348

Creates a report writing service that allows for the generation of a byte[] that contains a zip of files as well as the generation of a byte[] that contains a single file. Each file represents one device type and is filled with all devices associated with that device type.


BEIMA.Backend/MongoService/IMongoConnector.cs
```diff
@@ -106,6 +107,12 @@ public interface IMongoConnector
         public bool DeleteBuilding(ObjectId objectId);
 
         /// <summary>
+        /// Gets all devices that match the parameter filter's criteria
+        /// </summary>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        
+        public List<BsonDocument> GetFilteredDevices(FilterDefinition<BsonDocument> filter);
```
josephmoore99, line 115: I don't think you merged this in properly, the comments and the spacing for everything is off.

BEIMA.Backend/ReportService/ReportWriter.cs
```diff
@@ -0,0 +1,286 @@
...
+        /// property values are based on the keys in the parameter deviceType's fields and the values of the parameter device's fields.
+        /// </summary>
+        /// <param name="device"></param>
+        /// <param name="deviceType"></param>
+        /// <param name="delimiter"></param>
+        /// <returns></returns>
+        private static string GenerateDeviceValues(Device device, DeviceType deviceType, string delimiter = ",")
+        {
+            var values = new List<string>();
+            foreach (var deviceProp in typeof(Device).GetProperties())
```
josephmoore99, line 231: So I what you have here is probably working for now, but I think that we might not be able to guarantee that the values here are being placed in the same column as the previously generated column headers. I think it would be a good idea to also pass in the array of column headers (or the header string) to this method so that we can check to make sure that we guarantee the values in here are being placed into the correct columns, and we aren't mixing any of them up.

BEIMA.Backend.Test/ReportService/ReportServiceTest.cs
```diff
@@ -0,0 +1,435 @@
...
+                Assert.That(file2Rows.Count, Is.EqualTo(1));
+
+                // Tests that first row is a header row
+                Assert.That(GenerateDeviceHeaders(deviceTypeOne), Is.EqualTo(file1Rows[0]));
+                Assert.That(GenerateDeviceHeaders(deviceTypeTwo), Is.EqualTo(file2Rows[0]));
+            }
+        }
+
+        [Test]
+        public void DevicesAndDeviceTypesExists_GenerateAll_ZipContainesAllDevices()
```
AshlynAdamson, line 216:
```suggestion
        public void DevicesAndDeviceTypesExists_GenerateAll_ZipContainsAllDevices()
```

BEIMA.Backend/ReportService/ReportWriter.cs
```diff
@@ -0,0 +1,286 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BEIMA.Backend.ReportService
+{
+    /// <summary>
+    /// This class abstracts basic report generation operations. It is implemented as a 
+    /// singleton class. 
+    /// </summary>
+    public class ReportWriter : IReportService
```
josephmoore99, line 19: Having looked through this now, I think what you said about making this a static class sounds like it could be a good idea here.  My suggestion here however would be to have the devices/device types be passed in, as opposed to getting them from the database directly, and to do all the database stuff when we implement the endpoint for this. This way we don't have to make any mock databases when testing this. As it is right now, all of our database calls are on the endpoint classes themselves and I think it would be a good idea to make sure that we keep it that way to reduce dependencies.

BEIMA.Backend/ReportService/ReportWriter.cs
```diff
@@ -0,0 +1,286 @@
...
+        /// <summary>
+        /// Generates a byte[] containing a zip file with a file entry
+        /// per device type. Each of the file entries will contains
+        /// a header and then the data of all devices related to that specific
+        /// device type. If there are no device types returns then null is returned.
+        /// A file entry may contain only the header row if no device is associated
+        /// with that specific device type
+        /// </summary>
+        /// <returns></returns>
+        public byte[] GenerateAllDeviceReports()
```
AshlynAdamson, line 96: If we are wanting to export all devices, regardless of type, I think we should only print the mandatory fields of each device. I feel like the csv might get hard to read if some devices have some fields, while others don't, etc.

BEIMA.Backend.Test/ReportService/ReportServiceTest.cs
```diff
@@ -0,0 +1,435 @@
...
+            // Assert
+            Assert.That(bytes, Is.Not.Null);
+            Assert.DoesNotThrow(() => mockDb.Verify(mock => mock.GetDeviceType(It.Is<ObjectId>(oid => oid == deviceTypeOne.Id)), Times.Once));
+
+            var content = Encoding.UTF8.GetString(bytes);
+            var contentRows = content.Split(Environment.NewLine);
+            Assert.That(contentRows.Count, Is.EqualTo(3));
+
+            // Tests that first row is a header row
+            Assert.That(GenerateDeviceHeaders(deviceTypeOne), Is.EqualTo(contentRows[0]));
```
josephmoore99, line 135: I think copy/pasting the generate logic into the test here is not really helping us test that those methods actually worked. Instead, I would suggest testing against hard coded string values.

BEIMA.Backend.Test/ReportService/ReportServiceTest.cs
```diff
@@ -0,0 +1,435 @@
...
+            Assert.That(GenerateDeviceValues(deviceOne, deviceTypeOne), Is.EqualTo(contentRows[1]));
```
josephmoore99, line 136: I should note that the formatting for this is meant to be Assert.That(actual, Is.EqualTo(expected)), instead of the other way around. Otherwise the error message it produces on failures is a little confusing.

BEIMA.Backend/ReportService/ReportWriter.cs
josephmoore99: If we use a static class for this instead, we also don't need to use the interface/ReportDefinition classes, as we would be able to call this class directly. We only need those if we want to be able to mock this class in testing. However, because we aren't interacting with anything outside the program here (if we decide to move out the database stuff), there isn't really a reason why this would need to be mocked.

BEIMA.Backend/ReportService/ReportWriter.cs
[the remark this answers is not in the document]
josephmoore99: I think that could be a good idea, having the ability to generate a table based off only the mandatory fields so that we can put all of the devices onto it.

BEIMA.Backend.Test/ReportService/ReportServiceTest.cs
```diff
@@ -0,0 +1,435 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using BEIMA.Backend.MongoService;
+using Moq;
+using MongoDB.Bson;
+using System.IO.Compression;
+using System.IO;
+using MongoDB.Driver;
+
+namespace BEIMA.Backend.Test.ReportServicesa
```
josephmoore99, line 15:
```suggestion
namespace BEIMA.Backend.Test.ReportServices
```

BEIMA.Backend/ReportService/ReportWriter.cs
Kenny-Miller: Yeah I think it would also help remove the db dependency because it feels weird for this to be a service that relies on another service.

BEIMA.Backend/MongoService/IMongoConnector.cs
```diff
@@ -106,6 +107,12 @@ public interface IMongoConnector
         public bool DeleteBuilding(ObjectId objectId);
 
         /// <summary>
+        /// Gets all devices that match the parameter filter's criteria
+        /// </summary>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        public List<BsonDocument> GetFilteredDevices(FilterDefinition<BsonDocument> filter);
+
         /// Updates a building in the "buildings" collection, given a fully formed updated building.
```
josephmoore99, line 116:
```suggestion
        /// <summary>
        /// Updates a building in the "buildings" collection, given a fully formed updated building.
```

BEIMA.Backend/ReportService/ReportWriter.cs
```diff
@@ -0,0 +1,260 @@
...
+    public static class ReportWriter
+    {
+        /// <summary>
+        /// Generates a byte[] containing a file object. The
+        /// file contains a header based on the paramter deviceType
+        /// and then the data of all devices related to the specific deviceTypeId.
+        /// </summary>
+        /// <param name="deviceType"></param>
+        /// <param name="devices"></param>
+        /// <param name="delimiter"></param>
```
josephmoore99, line 24: Sorry to be that guy, but we needs ze param comments. Can just use "The device type", "The list of devices to enter", "the delimeter" etc.

BEIMA.Backend/ReportService/ReportWriter.cs
```diff
@@ -0,0 +1,260 @@
...
+        {
+            // Return if deviceType is null
+            if(deviceType == null)
+            {
+                return null;
+            }
+            // Ensure devices is not null;
+            if(devices == null)
+            {
+                devices = new List<Device>();
```
josephmoore99, line 36: Should we just return null at this point here too?

BEIMA.Backend.Test/ReportService/ReportServiceTest.cs
```diff
@@ -0,0 +1,464 @@
...
+using Moq;
+using MongoDB.Bson;
+using System.IO.Compression;
+using System.IO;
+using MongoDB.Driver;
+
+namespace BEIMA.Backend.Test.ReportServices
+{
+    [TestFixture]
+    public class ReportServiceTest : UnitTestBase
```
josephmoore99, line 18: Let's add a test to make sure that headers and values are aligned properly in their respective columns, just to be safe.

Kenny-Miller: image

These statements are testing that. We are able to take advantage of the fact that Lists guarantee that the order in which items are added are how they will appear.  I'm using sub lists just to help separate out the difference properties (general, fields, location, modified).

BEIMA.Backend/ReportService/ReportWriter.cs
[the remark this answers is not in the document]
Kenny-Miller: I will add this to a separate task in order to keep the pr size down

BEIMA.Backend.Test/ReportService/ReportServiceTest.cs
josephmoore99: I suppose I was thinking more along the lines of checking something like, check that column 1 has this header and this value, then check that column 2 has this header and this value etc. So basically checking column by column instead of row by row. But this is more of just a sanity check I think, these tests should probably be good enough.

BEIMA.Backend/ReportService/ReportWriter.cs
```diff
@@ -0,0 +1,275 @@
...
+                zipFileBytes = outStream.ToArray();
+            }
+            return zipFileBytes;
+        }
+
+        /// <summary>
+        /// Creates a list of strings filled with all of the property names in the Device object. The Fields
+        /// property is replaced with the field names contained in the parameter deviceType's Fields dictionary
+        /// </summary>
+        /// <param name="deviceType"></param>
```
josephmoore99, line 190: Missed a couple params here and in IMongoConnector

Kenny-Miller: Should be there now

josephmoore99 (approved):

thessbsu (approved):

roguedeleter (approved):

PR #379, opened 2022-04-01
User POST - AddUser

### LINKED ISSUE #302: Given a user, when the user POST endpoint is called, then the new user should be added to the database.
#272

### LINKED ISSUE #378: Research and add password hashing
#272 

Using only SHA to hash passwords is considered insecure and should not be done. This task will involve researching a more secure password hashing algorithm, and installing a well-known package that can easily handle password hashing for us. This will also involve implementing this when implementing the AddUser endpoint, since passwords will be hashed on the backend.

Research:

The BCrypt.Net package was chosen. This package uses the bcrypt algorithm, and it is generally accepted as a secure password hashing algorithm. This package was chosen because it is widely used, and it makes password hashing extremely easy as shown in the documentation here: https://github.com/BcryptNet/bcrypt.net

Using statement: 
`using BCryptNet = BCrypt.Net.BCrypt;`

To hash passwords:
`string passwordHash = BCryptNet.HashPassword(data.Password);`

To check a plain text password against a hashed password:
`bool passwordCorrect = BCrypt.Verify("plaintextPassword1!", passwordHash);`

Note on EnhancedHashPassword and EnhancedVerify methods in BCrypt.Net:

A conscious decision was made to not use the "Enhanced" versions of these methods. According to this article below, it is insecure to pre-hash passwords in the manner that EnhancedHashPassword does.

> An alternative approach is to pre-hash the user-supplied password with a fast algorithm such as SHA-256, and then to hash the resulting hash with bcrypt (i.e., bcrypt(base64(hmac-sha256(data:$password, key:$pepper)), $salt, $cost)). This is a dangerous (but common) practice that should be avoided due to [password shucking](https://www.youtube.com/watch?v=OQD3qDYMyYQ) and other issues when [combining bcrypt with other hash functions](https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html).

From: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

roguedeleter:
Closes #302 
Closes #378

This PR implements the AddUser endpoint. It also implements a password hashing package called BCrypt.Net. This uses the bcrypt algorithm, and it is generally accepted as a secure password hashing algorithm. This package was chosen because it is widely used, and it makes password hashing extremely easy as shown in the documentation here: https://github.com/BcryptNet/bcrypt.net

Using statement: 
`using BCryptNet = BCrypt.Net.BCrypt;`

To hash passwords:
`string passwordHash = BCryptNet.HashPassword(data.Password);`

To check a plain text password against a hashed password:
`bool passwordCorrect = BCrypt.Verify("plaintextPassword1!", passwordHash);`

Note on EnhancedHashPassword and EnhancedVerify methods in BCrypt.Net:

A conscious decision was made to not use the "Enhanced" versions of these methods. According to this article below, it is insecure to pre-hash passwords in the manner that EnhancedHashPassword does.

> An alternative approach is to pre-hash the user-supplied password with a fast algorithm such as SHA-256, and then to hash the resulting hash with bcrypt (i.e., bcrypt(base64(hmac-sha256(data:$password, key:$pepper)), $salt, $cost)). This is a dangerous (but common) practice that should be avoided due to [password shucking](https://www.youtube.com/watch?v=OQD3qDYMyYQ) and other issues when [combining bcrypt with other hash functions](https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html).

From: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

### DIFF
--- a/BEIMA.Backend.FT/TestObjects.cs
+++ b/BEIMA.Backend.FT/TestObjects.cs
@@ -162,6 +162,9 @@ namespace BEIMA.Backend.FT
 
             [JsonProperty(PropertyName = "role")]
             public string? Role { get; set; }
+
+            [JsonProperty(PropertyName = "lastModified")]
+            public LastModified? LastModified { get; set; }
         }
 
         public class Location
--- a/BEIMA.Backend.FT/TestObjects.cs
+++ b/BEIMA.Backend.FT/TestObjects.cs
@@ -143,6 +143,27 @@ namespace BEIMA.Backend.FT
             public LastModified? LastModified { get; set; }
         }
 
+        public class User
+        {
+            [JsonProperty(PropertyName = "_id")]
+            public string? Id { get; set; }
+
+            [JsonProperty(PropertyName = "username")]
+            public string? Username { get; set; }
+
+            [JsonProperty(PropertyName = "password")]
+            public string? Password { get; set; }
+
+            [JsonProperty(PropertyName = "firstName")]
+            public string? FirstName { get; set; }
+
+            [JsonProperty(PropertyName = "lastName")]
+            public string? LastName { get; set; }
+
+            [JsonProperty(PropertyName = "role")]
+            public string? Role { get; set; }
+        }
+
         public class Location
         {
             [JsonProperty(PropertyName = "latitude")]
--- a/BEIMA.Backend/UserFunctions/AddUser.cs
+++ b/BEIMA.Backend/UserFunctions/AddUser.cs
@@ -12,6 +12,7 @@ using MongoDB.Bson;
 using BEIMA.Backend.Models;
 using System.Net;
 using System.Web.Http;
+using BCryptNet = BCrypt.Net.BCrypt;
 
 namespace BEIMA.Backend.UserFunctions
 {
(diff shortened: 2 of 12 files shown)


BEIMA.Backend.FT/TestObjects.cs
```diff
@@ -143,6 +143,27 @@ public class Building
             public LastModified? LastModified { get; set; }
         }
 
+        public class User
```
josephmoore99, line 146: I think we also need a LastModified property here.

BEIMA.Backend/UserFunctions/AddUser.cs
```diff
@@ -0,0 +1,79 @@
...
+                                        passwordHash,
+                                        data.FirstName,
+                                        data.LastName,
+                                        data.Role);
+            }
+            catch (Exception)
+            {
+                return new BadRequestObjectResult(Resources.CouldNotParseBody);
+            }
+            // TODO: Use actual user.
```
AshlynAdamson, line 53: Will this TODO be done in another user task?

BEIMA.Backend.FT/TestObjects.cs
roguedeleter: Fixed 57c5fac

BEIMA.Backend/UserFunctions/AddUser.cs
roguedeleter: This TODO is part of all of our endpoints as of now. It is dependent on Kenny's work of adding authentication to endpoints. Once we have that, we will be able to insert the actual username of the person who is currently using the endpoint.

roguedeleter:

josephmoore99 (approved):

AshlynAdamson (approved):

Kenny-Miller (approved):

PR #380, opened 2022-04-01
Task#314 add device building integration

### LINKED ISSUE #314: Add Device page needs to have Building info integrated
References user story #82

Add device page needs to have a dropdown to select the building so the building ID can be added to the location data when saving to the DB

thessbsu:
Closes #314
Changed building input field to dropdown, added API call and error message if the call is not completed successfully

image
image
image


BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -243,7 +277,7 @@ const AddDevicePage = () => {
           <Form >
             <Row className={styles.buttonGroup}>
               <Col>
-                <FilledDropDown dropDownText={selectedDeviceType.name} items={deviceTypes} selectFunction={getFieldsForTypeId} buttonStyle={dropDownStyle} dropDownId={"typeDropDown"} />
+                <FilledDropDown dropDownText={selectedDeviceType.name} items={deviceTypes} selectFunction={getFieldsForTypeId} buttonStyle={deviceTypeDropDownStyle} dropDownId={"typeDropDown"} />
```
AshlynAdamson, line 280: Kind of noticed this now that we have buildings as well, but maybe we should add a label for the device type and building dropdown. Since the dropdown text gets changed when a selection is made, it might be nice to have a label. But also, would a user forget what that field is for after they select it? I could go both ways about a label

BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -157,9 +192,8 @@ const AddDevicePage = () => {
   * @return the compiled JSON
   */
   function createJSON(formFields){
-    //TODO temporary until we have error signaling figured out
     if (selectedDeviceType.name === 'Select Device Type'){
-      alert ('No device type selected');
+      Notifications.error("Device Type not selected", 'A device Type selection is required.');
```
roguedeleter, line 196: Capitalize "device"

BEIMA.Client/src/pages/Devices/AddDevicePage.js
thessbsu: I think it's pretty standard to have a label on it. I got used to seeing the "Select &lt;what the label would be&gt; text there before integrating the API calls I didn't really think about it.
Having one only has positive effects.

AshlynAdamson: True, we should get some added then

Kenny-Miller: An alternative to the label could be an option at the very top of the selection list that says Select Building. That way they see it every time they click the select box. We can then just make sure that if they try to add with that selected they get an error/warning

thessbsu: including the "select..." text in the list would require some tweaking of the way the list is generated to keep it, and the standard on the web is a label for it, and it's a much easier/quicker solution.

- [ ] add label to dropdowns for their purpose

thessbsu:
image

josephmoore99 (approved):

AshlynAdamson (approved):

PR #381, opened 2022-04-01
Task#159 add device empty field warning

### LINKED ISSUE #159: Add Device page warning for empty fields.
#157
We need to check if any of the fields are empty, and display a warning if they are and ask the user if they want to continue.

thessbsu:
Closes #159

Moved confirmation that device type is selected to happen before trying to create JSON object
Added async to createJSON so the warning can be ignored
added check for isConfirmed from warning message to determine if save should still be attempted

image

### DIFF
--- a/BEIMA.Client/src/pages/Devices/AddDevicePage.js
+++ b/BEIMA.Client/src/pages/Devices/AddDevicePage.js
@@ -230,7 +230,7 @@ const AddDevicePage = () => {
         let jsonKey = convertToDbFriendly(formName);
         
         if(formFields[i].value.length === 0){
-          warnings.push(`${formName} field is empty<br/>`);
+          warnings.push(`"${formName}" field is empty<br/>`);
         }
         
         //straight append when it's not part of a nested element
--- a/BEIMA.Client/src/pages/Devices/AddDevicePage.js
+++ b/BEIMA.Client/src/pages/Devices/AddDevicePage.js
@@ -165,7 +165,7 @@ const AddDevicePage = () => {
   */
   async function saveDeviceToDb(addButtonEvent) {
     if (selectedDeviceType.name === 'Select Device Type'){
-      let isConfirmed = await Notifications.error("Device Type not selected", 'A device Type selection is required.');
+      Notifications.error("Device Type not selected", 'A device Type selection is required.');
       return;
     }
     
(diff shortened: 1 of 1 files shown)


thessbsu:
- [ ] Change merge branch to main after branch Task#314... is merged since this is based on it

AshlynAdamson (dismissed):

BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -163,9 +163,14 @@ const AddDevicePage = () => {
   * gathers data from the form and saves the device to the DB
   * @param the Add Device button click event
   */
-  function saveDeviceToDb(addButtonEvent) {
+  async function saveDeviceToDb(addButtonEvent) {
+    if (selectedDeviceType.name === 'Select Device Type'){
+      Notifications.error("Device Type not selected", 'A device Type selection is required.');
```
roguedeleter, line 168: Capitalize "device"

BEIMA.Client/src/pages/Devices/AddDevicePage.js
```diff
@@ -228,6 +229,10 @@ const AddDevicePage = () => {
         let formJSON = '';
         let jsonKey = convertToDbFriendly(formName);
         
+        if(formFields[i].value.length === 0){
+          warnings.push(`${formName} field is empty<br/>`);
```
roguedeleter, line 233: Can we surround {formName} with quotes? And also end with a period.

So instead of:

>Name field is empty

Use:

>"Name" field is empty.

Kenny-Miller: Do we want to show every warning or just a generic some fields are empty message instead?

thessbsu: Quotes would be good.
As a user I'd prefer to be told exactly what I'm being warned about so I have an informed decision about clicking Proceed or Cancel. if we only mention that some fields are blank it would be user friendly to also highlight them (in yellow instead of red?) like we do with lat/lon so that (again) they know what they're being warned about.

- [ ] add quotes around field names

thessbsu: image

thessbsu:
image

Kenny-Miller (approved):

josephmoore99 (dismissed):

thessbsu:
- [ ] Check functionality after merge conflict resolution due to async addition and moving the device selection warning out of createJSON

josephmoore99 (approved):

roguedeleter (approved):

PR #385, opened 2022-04-02
377 Login Endpoint

### LINKED ISSUE #377: Login Endpoint
References user story #16

Create an endpoint that allows a username and password to be posted up. A user can
then be retrieved based on the username and password and then encoded into a jwt token. This
token is then returned. If a user is not found return a 401.

Kenny-Miller:
Closes #377

Implements a login endpoint that takes in a username and password, checks if a user has those credentials, and then returns a jwt token based on the found user. If the request parameters are invalid or the user doesn't exist, throws a 401 unauthorized.

### DIFF
--- a/BEIMA.Backend.FT/AuthFT.cs
+++ b/BEIMA.Backend.FT/AuthFT.cs
@@ -1,0 +1,84 @@
+﻿using JWT.Builder;
+using NUnit.Framework;
+using System.Net;
+using System.Threading.Tasks;
+using static BEIMA.Backend.FT.TestObjects;
+
+namespace BEIMA.Backend.FT
+{
+    [TestFixture]
+    public class AuthFT : FunctionalTestBase
+    {
+        [SetUp]
+        public async Task SetUp()
+        {
+            // Add in deleting all the users in the database once functionality implemented
+        }
+
+        [TestCase(null, null)]
+        [TestCase("valid", null)]
+        [TestCase(null, "valid")]
+        public void InvalidParameters_Login_ReturnsNull(string username, string password)
+        {
+            var loginRequest = new LoginRequest()
+            {
+                Username = username,
+                Password = password
+            };
+
+            var ex = Assert.ThrowsAsync<BeimaException>(async () =>
+                await TestClient.Login(loginRequest)
+            );
+            Assert.IsNotNull(ex);
+            Assert.That(ex?.StatusCode, Is.EqualTo(HttpStatusCode.Unauthorized));
+        }
+
+        [TestCase("notvalid", "notvalid")]
+        [TestCase("notvalid2", "notvalid2")]
+        public void ValidParametersNoUsers_Login_ReturnsNull(string username, string password)
+        {
+            var loginRequest = new LoginRequest()
+            {
+                Username = username,
+                Password = password
+            };
+
+            var ex = Assert.ThrowsAsync<BeimaException>(async () =>
+                await TestClient.Login(loginRequest)
+            );
+            Assert.IsNotNull(ex);
+            Assert.That(ex?.StatusCode, Is.EqualTo(HttpStatusCode.Unauthorized));
+        }
+
+        [TestCase("valid", "valid")]
+        [TestCase("valid2", "valid2")]
+        public async Task ValidParametersWithUsers_Login_ReturnsJwtToken(string username, string password)
+        {
+            var user = new User()
+            {
+                Username = username,
+                Password = password,
+                Role = "admin"
+            };
+
+            var addResponse = await TestClient.AddUser(user);
+            Assume.That(addResponse, Is.Not.Null);
+
+            var loginRequest = new LoginRequest()
+            {
+                Username = username,
+                Password = password
+            };
+
+            var token = await TestClient.Login(loginRequest);
+            Assert.That(token, Is.Not.Null);
+
+            var claims = new JwtBuilder().Decode<Claims>(token);
+            Assert.That(claims.Username, Is.EqualTo(username));
+            Assert.That(claims.Role, Is.EqualTo("admin"));
+            Assert.That(claims.Exp, Is.Not.Null);
+            Assert.That(claims.Sub, Is.EqualTo("User"));
+            Assert.That(claims.Iss, Is.EqualTo("Beima"));
+        }
+    }
+}
--- a/BEIMA.Backend.FT/AuthFT.cs
+++ b/BEIMA.Backend.FT/AuthFT.cs
@@ -19,7 +19,7 @@ namespace BEIMA.Backend.FT
         [TestCase(null, null)]
         [TestCase("valid", null)]
         [TestCase(null, "valid")]
-        public void InvalidParameters_Login_ReturnsNull(string username, string password)
+        public void InvalidParameters_Login_ReturnsHttpStatus401(string username, string password)
         {
             var loginRequest = new LoginRequest()
             {
@@ -36,7 +36,7 @@ namespace BEIMA.Backend.FT
 
         [TestCase("notvalid", "notvalid")]
         [TestCase("notvalid2", "notvalid2")]
-        public void ValidParametersNoUsers_Login_ReturnsNull(string username, string password)
+        public void ValidParametersNoUsers_Login_ReturnsHttpStatus401(string username, string password)
         {
             var loginRequest = new LoginRequest()
             {
--- a/BEIMA.Backend.FT/AuthFT.cs
+++ b/BEIMA.Backend.FT/AuthFT.cs
@@ -11,7 +11,7 @@ namespace BEIMA.Backend.FT
     public class AuthFT : FunctionalTestBase
     {
         [SetUp]
-        public async Task SetUp()
+        public void SetUp()
         {
             // Add in deleting all the users in the database once functionality implemented
         }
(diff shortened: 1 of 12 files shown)


BEIMA.Backend/AuthFunctions/Login.cs
```diff
@@ -0,0 +1,87 @@
...
+                }
+                password = data.Password;
+
+                // Get all users with request's username
+                var dbServce = MongoDefinition.MongoInstance;
+                var usernameFilter = MongoFilterGenerator.GetEqualsFilter("username", data.Username);
+                userDocs = dbServce.GetFilteredUsers(usernameFilter);
+            } catch (Exception)
+            {
+                return new UnauthorizedResult();
```
josephmoore99, line 56:
```suggestion
                return new UnauthorizedResult("Login unauthorized.");
```
We should make sure to return messages with our failures. Also, make sure to add the message to the resx file so that we can keep them all in one place like we are with the other ones.

BEIMA.Backend/AuthFunctions/Login.cs
```diff
@@ -0,0 +1,87 @@
...
+            
+            if(userDocs == null || userDocs.Count == 0)
+            {
+                return new UnauthorizedResult();
+            }
+
+            // Go through the users and find the one whose password hash matches the request's password
+            // Doesn't use filter for password as bcrypt hashing creates a unique salt on every hash.
+            User user = null;
+            for(var i = 0; i < userDocs.Count && user == null; i++)
```
josephmoore99, line 67: We should only need to look at a single users password. If were to implement it this way, I would think that if you used someone else's password, it would still let you log in. Usernames should be unique so we should be able to get the specific user based on that rather than trying to loop through a bunch of users off of the password here.

BEIMA.Backend/AuthFunctions/Login.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            List<BsonDocument> userDocs;
+            string password;
+            try
+            {
+                // Deserialize request
+                string requestBody = await new StreamReader(req.Body).ReadToEndAsync();
+                var data = JsonConvert.DeserializeObject<LoginRequest>(requestBody);
+                
+                if(data == null || data.Username == null || data.Password == null)
```
josephmoore99, line 44: Let's check for null or empty here

BEIMA.Backend/AuthFunctions/Login.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            if(user == null)
+            {
+                return new UnauthorizedResult();
+            }
+
+            // Create a jwt token
+            var authService = AuthenticationDefinition.AuthticationInstance;
+            var token = authService.CreateToken(user);
+            return new OkObjectResult(token);
```
josephmoore99, line 84: Should we instead try to return an object that looks like this?
{
"token" : "a234907soeifnlaksdnvliuzdhfklzk234jbfkzesbk"
}
I think just returning the token string should be fine, but was just wondering if we wanted to make it more obvious that this was a token.

BEIMA.Backend/MongoService/IMongoConnector.cs
```diff
@@ -121,6 +121,11 @@ public interface IMongoConnector
         public BsonDocument UpdateBuilding(BsonDocument doc);
 
         /// <summary>
+        /// Gets a list of User BsonDocuments using the passed in filter.
+        /// </summary>
+        /// <param name="filter">The filter to be applied.</param>
+        /// <returns>List of BsonDocuments</returns>
+        public List<BsonDocument> GetFilteredUsers(FilterDefinition<BsonDocument> filter);
         /// Inserts a user into the "users" collection
```
josephmoore99, line 129: Add newline here

BEIMA.Backend/AuthFunctions/Login.cs
Kenny-Miller: The userDocs is a prefiltered list gotten by the get equals filter that filters on username. It returns a list, but it should only contain 1 record once not allowing duplicates is fixed for the add user endpoint.

BEIMA.Backend/AuthFunctions/Login.cs
josephmoore99: We could also specify the time until expiration this way

BEIMA.Backend/AuthFunctions/Login.cs
Kenny-Miller: UnauthorizedResults do not have any constructor parameters. I can do this though instead: `return new BadRequestObjectResult(Resources.UnauthorizedMessage) { StatusCode = 401 };`

BEIMA.Backend.FT/AuthFT.cs
```diff
@@ -0,0 +1,88 @@
...
+            var token = await TestClient.Login(loginRequest);
+            Assert.That(token, Is.Not.Null);
+
+            var claims = new JwtBuilder().Decode<Claims>(token);
+            Assert.That(claims.Username, Is.EqualTo(username));
+            Assert.That(claims.Role, Is.EqualTo("admin"));
+
+            // Expiration of token is 7 days after creation. Should be 6 days 23 hours 59min xx seconds greater then datetime now
+            var nowPlus6 = DateTime.Now.AddDays(6).AddHours(23).AddMinutes(59).Ticks;
+            Assert.That(claims.Exp, Is.GreaterThan(nowPlus6));
```
josephmoore99, line 83: Could we also add a ```Is.LessThan(sevenDays + 1)``` or something?

BEIMA.Backend/AuthFunctions/Login.cs
```diff
@@ -0,0 +1,87 @@
...
+using BCryptNet = BCrypt.Net.BCrypt;
+using MongoDB.Bson;
+using System.Collections.Generic;
+
+namespace BEIMA.Backend.AuthFunctions
+{
+    /// <summary>
+    /// Handles a request to login
+    /// </summary>
+    public static class Login
```
josephmoore99, line 23: I think we should add a unit test class for this as well.

Kenny-Miller: Brain short circuited, will get that added.

BEIMA.Backend/AuthFunctions/Login.cs
josephmoore99: Let's tack on a .Single() linq method to that then so that we are only looking at one user rather than doing this.

BEIMA.Backend/AuthFunctions/Login.cs
josephmoore99: Let's do what Keelan did on his add user endpoint then. This is what he did:
image

BEIMA.Backend.Test/AuthFunctions/LoginTest.cs
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        [Test]
+        public async Task ValidParametersNoFilteredUser_Login_ReturnsNull()
+        {
+            // ARRANGE
+            Mock<IMongoConnector> mockDb = new Mock<IMongoConnector>();
+            mockDb.Setup(mock => mock.GetFilteredUsers(It.IsAny<FilterDefinition<BsonDocument>>()))
+                  .Returns(new List<BsonDocument>())
+                  .Verifiable();
```
josephmoore99, line 61: Don't forget to assign the mock db to MongoDefinition.MongoInstance

BEIMA.Backend.Test/AuthFunctions/LoginTest.cs
```diff
@@ -0,0 +1,165 @@
...
+
+namespace BEIMA.Backend.Test.AuthFunctions
+{
+    [TestFixture]
+    public class LoginTest : UnitTestBase
+    {
+        [Test]
+        public async Task NullPostBody_Login_ReturnsNull()
+        {
+            // ARRANGE
```
josephmoore99, line 25: Even if we don't expect the database to be hit in this test, we should still try to create a mock instance so that it doesn't try to do anything with the actual database if something goes wrong.

BEIMA.Backend.Test/AuthFunctions/LoginTest.cs
```diff
@@ -0,0 +1,165 @@
...
+                  .Verifiable();
+
+            var body = TestData._testValidKeysLoginRequest;
+            var request = CreateHttpRequest(RequestMethod.POST, body: body);
+            var logger = (new LoggerFactory()).CreateLogger("Testing");
+
+            // ACT
+            var token = ((ObjectResult)await Login.Run(request, logger)).Value?.ToString();
+
+            // ASSERT
```
josephmoore99, line 70: Don't forget to add the asserts to verify whether or not the mock database method ran.

josephmoore99 (approved):

AshlynAdamson (approved):

roguedeleter (approved):

AshlynAdamson (approved):

roguedeleter (approved):

BEIMA.Backend/AuthFunctions/Login.cs
Kenny-Miller: I think its fine just returning the string. The expiration date is also contained within the token.

josephmoore99 (approved):

PR #388, opened 2022-04-02
User JS service functions

### LINKED ISSUE #283: Create Get User and Get User List JS services
References user story #82 

We need javascript functions that call our get user and get user list endpoints and return the results from those endpoints.

### LINKED ISSUE #285: Create Add/Update/Delete User service JS functions
References user story #82

We need javascript functions that call our add/update/delete endpoints for a user.

AshlynAdamson:
Closes #283
Closes #285 

Currently, the add/get user endpoints are done, but these JS service functions are really simple to create. I got all of them created here, but we will need to wait to merge/test these when all the endpoints are done.

- [x] wait for rest of the user endpoints

### DIFF
--- a/BEIMA.Client/src/services/AddUser.js
+++ b/BEIMA.Client/src/services/AddUser.js
@@ -9,7 +9,7 @@ const API_URL = process.env.REACT_APP_API_URL;
  */
 export default async function addUser(userDetails) {
   //performs the post and returns an error message or the inserted user ID
-  const dbCall = await axios.post(API_URL + "user/", userDetails).catch(function (error) {
+  const dbCall = await axios.post(API_URL + "user", userDetails).catch(function (error) {
       if (error.response) {
         return error.response;
     }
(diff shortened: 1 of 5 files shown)


BEIMA.Client/src/services/AddUser.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * POSTs a JSON object to the API for storage in DB
+ *
+ * @param The user details to add to the DB
+ * @return Error message or the inserted user ID
+ */
+export default async function addUser(userDetails) {
+  //performs the post and returns an error message or the inserted user ID
+  const dbCall = await axios.post(API_URL + "user/", userDetails).catch(function (error) {
```
josephmoore99, line 12:
```suggestion
  const dbCall = await axios.post(API_URL + "user", userDetails).catch(function (error) {
```

josephmoore99 (approved):
I think it should be fine for us to merge this in before implementing all of the user endpoints, just as long as we don't actually use/test against them until they have been implemented.

Kenny-Miller (approved):

roguedeleter (approved):

PR #391, opened 2022-04-03
ReportWriter: Device Type Report

### LINKED ISSUE #389: Create DeviceType CSV Report
References user story #275 

Add a method to the report writer to be able to generate a report for all the device types. It should be a single file that takes in a list of all the device types and all the devices. It should then print out all of the data associated with the device type except for their fields as well as a count of all devices associated with them.

Kenny-Miller:
Closes #389

Adds a method to the report writer to allow it to generate a file containing a list of
all device types in the database as well as a count of how many devices are associated
with each device type.

### DIFF
--- a/BEIMA.Backend/ReportService/ReportWriter.cs
+++ b/BEIMA.Backend/ReportService/ReportWriter.cs
@@ -31,11 +31,6 @@ namespace BEIMA.Backend.ReportService
                 return null;
             }
 
-            if(devices == null || devices.Count == 0)
-            {
-                return null;
-            }
-
             // Create dict to group devices based on device type id
             var typeToDevices = new Dictionary<ObjectId, List<Device>>();
             foreach (var deviceType in deviceTypes)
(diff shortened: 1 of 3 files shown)


BEIMA.Backend/ReportService/ReportWriter.cs
```diff
@@ -14,6 +14,76 @@ namespace BEIMA.Backend.ReportService
...
+        /// <param name="delimiter">String used to seperate every value in the report</param>
+        /// <returns>Byte[] containing a file filled with devicetype data and the number of devices associated with that devicetype</returns>
+        public static byte[] GenerateDeviceTypeReport(List<DeviceType> deviceTypes, List<Device> devices, string delimiter = ",")
+        {
+            if(deviceTypes == null || deviceTypes.Count == 0)
+            {
+                return null;
+            }
+
+            if(devices == null || devices.Count == 0)
```
josephmoore99, line 34: I feel like since we are generating reports specifically for device types, I think it's ok to make device type reports that have no devices. I'm open to discussion on this.

BEIMA.Backend/ReportService/ReportWriter.cs
```diff
@@ -183,6 +247,68 @@ public static byte[] GenerateAllDeviceReports(List<DeviceType> deviceTypes, List
...
+
+        /// <summary>
+        /// Creates a list of strings filled with all of the property values in a DeviceType object. In addition
+        /// another value is added for the count of the number of devices associated with the DeviceType.
+        /// </summary>
+        /// <param name="deviceType">DeviceType that the values should be based on</param>
+        /// <param name="associatedDevices">List of devices associated with the DeviceType</param>
+        /// <returns></returns>
+        private static List<string> GenerateDeviceTypeReportValues(DeviceType deviceType, List<Device> associatedDevices)
+        {
```
josephmoore99, line 282: Feels bad not including the custom fields. Maybe we could put all of the custom fields into a single cell for each device type? Idk, just feels like we should be including them.

BEIMA.Backend/ReportService/ReportWriter.cs
Kenny-Miller: I'm good with that as well.

BEIMA.Backend/ReportService/ReportWriter.cs
Kenny-Miller: I'm somewhat hesitant to do so just because csv's whose rows don't have the same amount of columns tend to be pretty messy and hard to look at.

josephmoore99: I was thinking that we would use a different dilimeter than a comma for that cell, so that they are all in the same box, but I could see that being a bit messy as well. I'm just wondering if there is any reasonable way for us to include them.

josephmoore99 (approved):

roguedeleter (approved):

PR #393, opened 2022-04-03
User service integration into User pages

### LINKED ISSUE #284: Update User pages with appropriate api calls
References user story #82

We need our static user pages to use the user service functions to get data from the database and store it on the page.

AshlynAdamson:
Closes #284 

I've integrated the User service functions into their appropriate pages. Right now, the add user works, we will need to wait for the rest of the endpoints to confirm the integrations works. Wanted to get it up for review though for any possible feedback. I am confident that there won't be much more to change once the endpoints are done. I will also post a demo video once the endpoints are done.

- [x] wait for rest of user endpoints

### DIFF
--- a/BEIMA.Client/src/pages/Users/AddUserPage.js
+++ b/BEIMA.Client/src/pages/Users/AddUserPage.js
@@ -4,6 +4,9 @@ import { IoArrowBack } from "react-icons/io5";
 import { Card, Button, Form, Row, Col } from 'react-bootstrap';
 import FormListWithErrorFeedback from '../../shared/FormList/FormListWithErrorFeedback.js';
 import { useEffect, useState } from "react";
+import AddUser from '../../services/AddUser.js';
+import * as Constants from '../../Constants.js';
+import * as Notifications from '../../shared/Notifications/Notification.js';
 
 const AddUserPage = () => {
   const mandatoryUserFields = {
@@ -17,7 +20,6 @@ const AddUserPage = () => {
   
   const [userFields, setUserFields] = useState(mandatoryUserFields);
   const [setPageName] = useOutletContext();
-  const [fullUserJSON, setFullUserJSON] = useState({});
   const [errors, setErrors] = useState({});
   
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


BEIMA.Client/src/pages/Users/UserPage.js
```diff
@@ -91,13 +76,26 @@ const UserPage = () => {
...
-      alert("you sure?");
+    const deleteUserCall = async () => {
+      let deleteNotif = await Notifications.warning("Warning: User Deletion", [`Are you sure you want to delete User ${username}?`]);
+      if(deleteNotif.isConfirmed){
+        let deleteResult = await DeleteUser(userId);
+        if(deleteResult.status === 200){
+          Notifications.success("User Deletion Successful", `User ${username} successfully deleted.`);
+          navigate('/users');
+        } else {
+          Notifications.error("Unable to Delete User", `Deletion of User ${username} failed.`);
```
AshlynAdamson, line 96: I can update this to use the message from the endpoint in the case of an admin is trying to delete the only admin account

BEIMA.Client/src/pages/Users/UserPage.js
```diff
@@ -91,13 +76,26 @@ const UserPage = () => {
...
+        setEditable(false);
+      } else {
+        Notifications.error("Unable to Update User", `Update of User ${username} failed.`);
+      }
     }
     
-    const deleteUserCall = () => {
-      //TODO integrate with styled error/confirm message
-      alert("you sure?");
+    const deleteUserCall = async () => {
```
AshlynAdamson, line 88: I think having a check for if the current user is trying to delete themselves would be good. I can update this as I work on the authentication side of things

BEIMA.Client/src/pages/Users/UserPage.js
josephmoore99: I think that would be a good idea, just to be more specific about the error. We could even make a bunch of else ifs instead if we go based off of the status code and/or message and customize our messages for each of those. But yeah, just relaying the message in the response would be the simplest way to go.

BEIMA.Client/src/pages/Users/AddUserPage.js
```diff
@@ -37,27 +39,46 @@ const AddUserPage = () => {
...
     }
     
     if((!('Password' in passwordErrors) && !('Password Confirmation' in passwordErrors)) &&
        (userFields["Password"] !== userFields["Password Confirmation"])
       ) {
-         passwordErrors["Password"] = 'Paswords do not match';
-         passwordErrors["Password Confirmation"] = 'Paswords do not match';
+         passwordErrors["Password"] = 'Passwords do not match';
+         passwordErrors["Password Confirmation"] = 'Passwords do not match';
     }
```
josephmoore99, line 55: Let's add in checks for an 8 character password length and require special/number/uppercase characters as well

BEIMA.Client/src/pages/Users/UserPage.js
josephmoore99: I think we should definitely prompt the user before deleting their own account, but I don't know that we should prevent them from doing so.

AshlynAdamson:
Here's that demo

https://user-images.githubusercontent.com/58532471/161454200-01ba8183-f050-4d24-b443-a629047e4d56.mp4

josephmoore99 (approved):

BEIMA.Client/src/pages/Users/AddUserPage.js
```diff
@@ -35,29 +42,53 @@ const AddUserPage = () => {
   
   function checkPassword(){
     let passwordErrors = {};
+    let requirements = new RegExp("^(?=.*[a-z])(?=.*[A-Z])(?=.*\\d)(?=.*[-+_!@#$%^&*.,?]).+$");
```
roguedeleter, line 45: Should this be in a constants file?

AshlynAdamson: Yeah, that's probably a good place for it

AshlynAdamson: Added it to the constants file

roguedeleter (approved):

josephmoore99 (approved):

PR #398, opened 2022-04-03
Add backend password requirement validation

### LINKED ISSUE #395: Add password validation to user endpoints
References user story #274

Add password validation on the user endpoint. A password should have at least 8 characters, a number, an uppercase character, and a special character.

josephmoore99:
Closes #395 

This adds validation for passwords on the backend to make sure they fulfill our requirements for passwords.

### DIFF
--- a/BEIMA.Backend/UserFunctions/UpdateUser.cs
+++ b/BEIMA.Backend/UserFunctions/UpdateUser.cs
@@ -13,6 +13,7 @@ using MongoDB.Bson.Serialization;
 using BEIMA.Backend.Models;
 using System.Net;
 using BCryptNet = BCrypt.Net.BCrypt;
+using System.Linq;
 
 namespace BEIMA.Backend.UserFunctions
 {
@@ -67,12 +68,20 @@ namespace BEIMA.Backend.UserFunctions
                 }
                 else
                 {
-                    // TODO: Add validation for password length and special characters
+                    // Verify password meets all requirements
+                    if (string.IsNullOrEmpty(updatedUserRecord.Password) ||
+                       updatedUserRecord.Password.Length < 8 ||
+                       !updatedUserRecord.Password.Any(c => char.IsDigit(c)) ||
+                       !updatedUserRecord.Password.Any(c => char.IsUpper(c)) ||
+                       !updatedUserRecord.Password.Any(c => !char.IsLetterOrDigit(c)))
+                    {
+                        return new BadRequestObjectResult(Resources.InvalidPasswordMessage);
+                    }
                     updatedUserRecord.Password = BCryptNet.HashPassword(updatedUserRecord.Password);
                 }
 
                 // Set user properties to new values
-                user = new User(new ObjectId(id), updatedUserRecord.Username, updatedUserRecord.Password, 
+                user = new User(new ObjectId(id), updatedUserRecord.Username, updatedUserRecord.Password,
                     updatedUserRecord.FirstName, updatedUserRecord.LastName, updatedUserRecord.Role);
             }
             catch (Exception)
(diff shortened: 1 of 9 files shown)


BEIMA.Backend/UserFunctions/UpdateUser.cs
```diff
@@ -67,12 +68,20 @@ public static class UpdateUser
                 }
                 else
                 {
-                    // TODO: Add validation for password length and special characters
+                    // Verify password meets all requirements
+                    if (string.IsNullOrEmpty(updatedUserRecord.Password) ||
+                       updatedUserRecord.Password.Length < 8 ||
+                       !updatedUserRecord.Password.Any(c => char.IsDigit(c)) ||
+                       !updatedUserRecord.Password.Any(c => char.IsUpper(c)) ||
```
roguedeleter, line 75: Also need to verify that the password also contains lower case characters as well, otherwise a user could use an all upper case password

roguedeleter: Maybe we should use a regex instead of LINQ expressions?

roguedeleter: Maybe something like this:

`^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9]).{8,}$`

BEIMA.Backend/UserFunctions/UpdateUser.cs
```diff
@@ -67,12 +68,20 @@ public static class UpdateUser
                 }
                 else
                 {
-                    // TODO: Add validation for password length and special characters
+                    // Verify password meets all requirements
+                    if (string.IsNullOrEmpty(updatedUserRecord.Password) ||
```
roguedeleter, line 72: For UpdateUser, it is okay for password to be null or empty. They are not required to enter password if all they want to do is change other properties like FirstName/LastName, etc

BEIMA.Backend/UserFunctions/UpdateUser.cs
roguedeleter: Actually, ignore the above. The above regex does not require a special character. Try this instead:

`^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{8,}$`

BEIMA.Backend/UserFunctions/UpdateUser.cs
roguedeleter: Oh wait, I see that this is handled in the if/else block surrounding it. GitHub was trimming off that part and I couldn't see it. Regardless, the "string.IsNullOrEmpty" here is redundant since the if block already checks for that.

josephmoore99: This still allows for empty/null passwords on update. We can only get into this else statement if the password is not null or empty.

BEIMA.Backend/UserFunctions/UpdateUser.cs
josephmoore99: Should we also include other characters like colons and slashes and stuff?

BEIMA.Backend/UserFunctions/UpdateUser.cs
josephmoore99: Fixed ```313159c```

BEIMA.Backend/UserFunctions/UpdateUser.cs
roguedeleter: We can go either way with it, what are your thoughts on it?

roguedeleter (dismissed):

AshlynAdamson (approved):

roguedeleter (approved):

AshlynAdamson (approved):

PR #435, opened 2022-04-08
Docker implementation

### LINKED ISSUE #374: research docker implementation/get containers set up
References user story #434

- [x] begin research on setting up project in Docker container(s)
- [x] initial docker setup (containers launch)

thessbsu:
Closes #374

Need docker and docker-compose installed
"sudo docker-compose up" should build and deploy the containers
"sudo docker-compose down" should shut them all down

once they're running you'll need the IP of the container to get to the UI, or to attempt an API call with postman
sudo docker inspect beima_&lt;container&gt;_1 | grep \"IPAddress\"
&lt;container&gt; being: web, api, mongo, minio

Moved Cypress to dev dependencies so that it's not pulled/built with the web container


BEIMA.Backend/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM mcr.microsoft.com/dotnet/sdk:6.0 AS installer-env
+EXPOSE 7071
+# Build requires 3.1 SDK
+COPY --from=mcr.microsoft.com/dotnet/core/sdk:3.1 /usr/share/dotnet /usr/share/dotnet
```
josephmoore99, line 4: Why do we need 3.1?

thessbsu: should it be different? this is the 6th or 7th iteration of tryin to get this one to work, this was the config that finally did

josephmoore99: I believe we have been using .net 6.0 if I'm not mistaken

thessbsu: It's the SDK/Framework version vs the C# version.
C# is at version 6, .NET Core is the cross platform version of the .NET framework whereas the standard .NET Framework is Windows only.

As far as I'm aware, without .NET Core our app won't run.

josephmoore99: I believe that what is now just "dotnet" used to be "dotnet core". You are correct in that it used to be the case that dotnet core was used for multiplatform environments and that framework was used for Windows. However, my understanding was that after dotnet core 3.1 and some framework version, they stopped splitting releases into dotnet core and framework and just made one "dotnet".

josephmoore99: Does it not work if you don't include dotnet Core 3.1?

thessbsu: Just tried it without the copy and in the MS provided container it does!
I was thinking it would still be required since in one of the iterations (1 -> 5 or so) when I was building it on top of the ubuntu 20.04 image manually/directly it complained about the dotnet command not being there until the SDK was added too.

Looks like in the MS provided container there's enough already installed to do it without it.

josephmoore99 (dismissed):
Not super familiar with docker stuff, but as long as it's all working, it looks good to me. Thanks for getting this done!

thessbsu:
Me neither for sure. at this point it's just the Draft that Ashlyn asked for. it's the proof of concept that "it builds" tomorrow I'll be looking to see how they communicate with each other and getting it to DO something.

Right now the main test is "can someone pull the branch, install docker and docker compose, and actually build it"

It should work on any OS since it's independent, but if there's some other config that's needed that I missed, or something I'd rather know now as she suggested than get 4 more levels deep and figure out there's something that breaks the downstream that I've put up.

Storage mapping, port mapping, etc is a little up in the air at the moment

BEIMA.Client/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+# Create directory for front end
+RUN mkdir /beima
+# Set work directory
+WORKDIR /beima
+# Update the apt package manager
+RUN apt update
+# Set time zone env variable and install tzdata package using it
+RUN TZ=Etc/UTC apt-get -y install tzdata
+# Move all required folders into the container
+COPY ./BEIMA.Client/ /beima/BEIMA.Client/
```
AshlynAdamson, line 16: We will need to make sure Dr. Henderson gets the local.settings.json and the .env added to our project before running this. It's been a while since I played with Docker, I'm glad we can copy over what we need and won't need to worry about getting those files added a different way

AshlynAdamson:
I was able to get this to work using docker desktop. I did notice the api is using the azure backend app, which I thought was odd since I have my local.settings.json to point to my local stuff. But good work so far!

BEIMA.Backend/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM mcr.microsoft.com/dotnet/sdk:6.0 AS installer-env
+EXPOSE 7071
+# Copy in source
+COPY ./BEIMA.Backend/ /src/dotnet-function-app
+# set up for publish
+RUN cd /src/dotnet-function-app && mkdir -p /home/site/wwwroot && dotnet publish *.csproj --output /home/site/wwwroot
+# To enable ssh & remote debugging on app service change the base image to the one below
+#FROM mcr.microsoft.com/azure-functions/dotnet:4-appservice
+FROM mcr.microsoft.com/azure-functions/dotnet:4
+ENV AzureWebJobsScriptRoot=/home/site/wwwroot AzureFunctionsJobHost__Logging__Console__IsEnabled=true
```
AshlynAdamson, line 10: I don't think we want this to hit our azure backend, but instead be local. I know this is still a work in progress, thought I'd bring it up

BEIMA.Client/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+# Install react dom
+RUN npm install react react-dom --save
+# Build client in production mode
+RUN npm run build
+# Install web server
+RUN npm install -g serve
+# Make sure container is in the correct directory to serve
+WORKDIR /beima/BEIMA.Client
+# Run the front end
+CMD ["serve", "-s", "build"]
```
AshlynAdamson, line 48: I'm seeing the CORS issue again, I found this article that might help https://maximillianxavier.medium.com/solving-cors-problem-on-local-development-with-docker-4d4a25cd8cfe

BEIMA.Client/Dockerfile
thessbsu: I was thinking about that too. There's likely going to have to be a production version of at least the .env file to point at the DNS name of the api on the local docker network instead of localhost.

BEIMA.Client/Dockerfile
thessbsu: this shouldn't be too hard, just another minimal container, then point the api calls at it, and it at the api container... I think

thessbsu: looks like it's going to work, just need to work out a config issue

AshlynAdamson:
When running all the docker stuff, I sometimes don't get port numbers for the web client or the db. And when I do, they are not our usual port numbers. Thought I'd bring it up

Kenny-Miller:
> When running all the docker stuff, I sometimes don't get port numbers for the web client or the db. And when I do, they are not our usual port numbers. Thought I'd bring it up

We'll probably want to talk with Professor Henderson to see if we can get a range of ports that only our team can use on the sdp server.

thessbsu:
Backend docker builds the API, I was able to get results inside the actual container via curl on localhost, accessing with postman on the external docker IP (the 172.*.*.* address)  also yielded results. The console reports it's listening on localhost, but it appears to be listening on all interfaces

BEIMA.Backend/Dockerfile
thessbsu: building from a base image and adding our components now

BEIMA.Client/Dockerfile
thessbsu: production version/flags will be needed for local.settings.json, looks like the env file will be fine

BEIMA.Client/Dockerfile
thessbsu: no proxy needed with custom network defined and IPs

thessbsu:
ports from localhost are mapped to the containers, if they need to be changed, or the custom network needs to be changed it can all be done from the docker-compose file

until a "deploy" environment for CurrentEnv is available
Local.settings.json changes
  MinioBucket should be "files"
  LocalMongoConnection in the local.settings.json needs to be changed to "mongodb://192.168.0.2:27017"
modify BEIMA.Backend\StorageService\MinioStorageProvider.cs with:
            client = new MinioClient().WithCredentials(accessKey, secretKey).WithEndpoint("192.168.0.4",9000);

BEIMA.Backend/Dockerfile
thessbsu: build for the API was completely changed, but SDK 6 was required

josephmoore99 (dismissed):

Kenny-Miller (approved):

thessbsu:
approvals can wait on this one until it's converted to a real PR from Draft. It's not ready to merge until then, and I don't know if it will dismiss approvals when I convert it either

AshlynAdamson (approved):

PR #439, opened 2022-04-10
#405 Auth Device Endpoints

### LINKED ISSUE #405: Adding Authentication to Device Endpoints
References user story #16

Add checking for an authentication token before running any device endpoints. If the token is invalid or doesn't exist, throw a 401.
If token is valid continue on with the request

In addition update the last modified user value to use the username in the claims object that is gotten from the auth token.

Kenny-Miller:
Closes #405

Add authentication in front of all the device endpoints. Also updated the unit and functional tests to handle this change.

### DIFF
--- a/BEIMA.Backend.FT/BeimaClient.cs
+++ b/BEIMA.Backend.FT/BeimaClient.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Text;
 using System.Threading.Tasks;
 using static BEIMA.Backend.FT.TestObjects;
@@ -71,7 +72,7 @@ namespace BEIMA.Backend.FT
         /// <param name="body">The object to send with the request.</param>
         /// <returns>The http response message of the request.</returns>
         /// <exception cref="HttpRequestException"></exception>
-        public async Task<HttpResponseMessage> SendRequest(string route, HttpVerb verb, object? body = null, string queryString = "")
+        public async Task<HttpResponseMessage> SendRequest(string route, HttpVerb verb, object? body = null, string queryString = "", string? token = null)
         {
             StringContent? jsonString = null;
             HttpResponseMessage response;
@@ -97,10 +98,21 @@ namespace BEIMA.Backend.FT
             switch (verb)
             {
                 case HttpVerb.GET:
-                    response = await _httpClient.GetAsync(route);
+                    var getRequest = new HttpRequestMessage(HttpMethod.Get, route);
+                    if(token != null)
+                    {
+                        getRequest.Headers.Authorization = new AuthenticationHeaderValue("Bearer", token);
+                    }
+                    response = await _httpClient.SendAsync(getRequest);
                     break;
                 case HttpVerb.POST:
-                    response = await _httpClient.PostAsync(route, jsonString);
+                    var postRequest = new HttpRequestMessage(HttpMethod.Post, route);
+                    if (token != null)
+                    {
+                        postRequest.Headers.Authorization = new AuthenticationHeaderValue("Bearer", token);
+                    }
+                    postRequest.Content = jsonString;
+                    response = await _httpClient.SendAsync(postRequest);
                     break;
                 default:
                     throw new HttpRequestException($"Invalid request verb: {verb}.");
@@ -127,7 +139,7 @@ namespace BEIMA.Backend.FT
         /// <param name="files">List of files to send with the request</param>
         /// <returns>The http response message of the request.</returns>
         /// <exception cref="HttpRequestException"></exception>
-        public async Task<HttpResponseMessage> SendMultiPartRequest(string route, Object data, FormFileCollection? files = null, string? queryString = "")
+        public async Task<HttpResponseMessage> SendMultiPartRequest(string route, Object data, FormFileCollection? files = null, string? queryString = "", string? token = null)
         {
             StringContent json;
             HttpResponseMessage response;
@@ -158,7 +170,14 @@ namespace BEIMA.Backend.FT
                 }
             }
 
-            response = await _httpClient.PostAsync(route, form);
+
+            var postRequest = new HttpRequestMessage(HttpMethod.Post, route);
+            if (token != null)
+            {
+                postRequest.Headers.Authorization = new AuthenticationHeaderValue("Bearer", token);
+            }
+            postRequest.Content = form;
+            response = await _httpClient.SendAsync(postRequest);
 
             try
             {
@@ -180,9 +199,9 @@ namespace BEIMA.Backend.FT
         /// </summary>
         /// <param name="id">The id of the device.</param>
         /// <returns>The device with the given id.</returns>
-        public async Task<Device> GetDevice(string id)
+        public async Task<Device> GetDevice(string id, string? token = null)
         {
-            var response = await SendRequest($"api/device/{id}", HttpVerb.GET);
+            var response = await SendRequest($"api/device/{id}", HttpVerb.GET, token: token);
             var content = await response.Content.ReadAsStringAsync();
             return JsonConvert.DeserializeObject<Device>(content);
         }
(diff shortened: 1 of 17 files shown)


BEIMA.Backend/DeviceFunctions/GetDevice.cs
```diff
@@ -32,49 +33,49 @@ public static class GetDevice
         {
             log.LogInformation("C# HTTP trigger function processed a device get request.");
 
-            // Process as device GET request for retrieving device information.
-            if (string.Equals(req.Method, "get", StringComparison.OrdinalIgnoreCase))
```
Kenny-Miller, line 36: Not sure why the file is showing everything as changed. All that was removed was line 36 where we check if the request is a get request as that is already handled by the Http Trigger

BEIMA.Backend.FT/BeimaClient.cs
```diff
@@ -97,10 +98,21 @@ public async Task<HttpResponseMessage> SendRequest(string route, HttpVerb verb,
             switch (verb)
             {
                 case HttpVerb.GET:
-                    response = await _httpClient.GetAsync(route);
```
josephmoore99, line 100: Instead of doing this and creating a new AuthenticationHeaderValue each time we send a request, I would suggest just setting the _httpClient authorization header on the login method:
```httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", "Your Oauth token");```
This way we remove a lot of complexity in having to pass in our token every time we make a request, and just store it in the client object's header itself. We can just login and not have to worry about the token in any of the FTs.

File: BEIMA.Backend.FT/DeviceFT.cs [hunk 3]
```diff
@@ -82,25 +90,38 @@ public async Task OneTimeSetUp()
...
                 {
-                    await TestClient.DeleteDevice(device.Id);
+                    await TestClient.DeleteDevice(device.Id, _authToken);
                 }
             }
         }
 
         [TestCase("xxx")]
         [TestCase("1234")]
         [TestCase("1234567890abcdef1234567x")]
```
josephmoore99, line 105: We might as well include some test cases with some validly formatted id's here as well.

BEIMA.Backend.FT/FunctionalTestBase.cs
```diff
@@ -29,6 +34,21 @@ public void BaseOneTimeSetUp()
             // Uncomment for Azure server:
             // BeimaClient Client = new BeimaClient("https://beima-service.azurewebsites.net");
             TestClient = new BeimaClient("http://localhost:7071");
+
+            // Have at least one admin user in the DB at all times as tests wil need it for auth
+            var initialUserList = await TestClient.GetUserList();
```
josephmoore99, line 39: This isn't going to work because GetUserList and AddUser are going to need to require a logged in admin user to even use (once we get that implemented). This might be ok for now but we will need another solution when we put authentication on the user endpoints.

BEIMA.Backend.FT/BeimaClient.cs
Kenny-Miller: That's how I originally had it, but wasn't sure if I should leave it off so that we could have ft tests to check sending without a token. Since we have the mock tests though its probably fine.

BEIMA.Backend/DeviceFunctions/AddDevice.cs
```diff
@@ -34,6 +35,14 @@ public static class AddDevice
             log.LogInformation("C# HTTP trigger function processed a device post request.");
 
 
+            var authService = AuthenticationDefinition.AuthticationInstance;
+            var claims = authService.ParseToken(req);
+
+            if(claims == null)
+            {
+                return new ObjectResult(Resources.UnauthorizedMessage) { StatusCode = 401 };
```
josephmoore99, line 43: If we aren't going to include a response message, we might as well just use an UnauthorizedResult object instead:
```return new UnauthorizedResult();```

File: BEIMA.Backend/DeviceFunctions/AddDevice.cs [hunk 6]
```diff
@@ -34,6 +35,14 @@ public static class AddDevice
             log.LogInformation("C# HTTP trigger function processed a device post request.");
 
 
+            var authService = AuthenticationDefinition.AuthticationInstance;
+            var claims = authService.ParseToken(req);
+
+            if(claims == null)
```
josephmoore99, line 41:
```suggestion
            if(AuthenticationDefinition.AuthenticationInstance.ParseToken(req) == null)
```
I think it would be ideal if we could condense this as much as possible, since we are going to have to put this on every endpoint.

BEIMA.Backend/DeviceFunctions/AddDevice.cs
```diff
@@ -34,6 +35,14 @@ public static class AddDevice
             log.LogInformation("C# HTTP trigger function processed a device post request.");
 
 
+            var authService = AuthenticationDefinition.AuthticationInstance;
```
josephmoore99, line 38:
```suggestion
            var authService = AuthenticationDefinition.AuthenticationInstance;
```
There is a spelling mistake on the class property, we should probs fix that.

BEIMA.Backend/DeviceFunctions/AddDevice.cs
josephmoore99: nvm, I see that we have to use claims.Username for the LastModified field below.

BEIMA.Backend/DeviceFunctions/GetDevice.cs
josephmoore99: It's probably because the tabbing was changed.

BEIMA.Backend.FT/FunctionalTestBase.cs
josephmoore99: Unless of course we modify the user endpoints to not check for authentication until an admin user exists in the database. Was that what you were planning?

BEIMA.Backend.FT/BeimaClient.cs
josephmoore99: We could create a new instance of the BeimaClient to test that instead of using TestClient on those tests.

josephmoore99: Or implement logout functionality?

BEIMA.Backend/DeviceFunctions/AddDevice.cs
Kenny-Miller: The unauthorized message gets returned as a string though. I thought that was the point of doing it this way instead of returning `UnauthorizedResult();`

BEIMA.Backend.FT/FunctionalTestBase.cs
Kenny-Miller: I think for the testing side of things, we might end up doing a raw-mongo connecter call to add a user in. I think this will be ok just because for the live application there will need to be at least on user in the db as you'll have to login before doing anything. Keelan, Ashlyn, and I were kinda discussing this issue and possible solutions, and were gonna bring it up next scrum.

BEIMA.Backend.FT/BeimaClient.cs
Kenny-Miller: I like the idea of separate clients. I'll go with that one

BEIMA.Backend/DeviceFunctions/AddDevice.cs
josephmoore99: Right nvm, I didn't notice the Resources.UnauthorizedMessage you passed in, ignore me lol

BEIMA.Backend.FT/BeimaClient.cs
```diff
@@ -63,6 +64,16 @@ public BeimaClient(string baseUrl)
             _httpClient.DefaultRequestHeaders.Add("Accept", "application/json");
         }
 
+        /// <summary>
+        /// Sets the authentication header of the http client to 'Bearer token'
+        /// for all outgoing http requests
+        /// </summary>
+        /// <param name="token">Jwt Token</param>
+        public void SetAuthenticationHeader(string token)
```
josephmoore99, line 72: I was thinking that we could even just set this in the login method itself, but this works just fine too.

BEIMA.Backend.FT/DeviceFT.cs
Kenny-Miller: Went ahead and added a unauthorized test for every normal test

josephmoore99 (approved):

AshlynAdamson (approved):

Issue #370, opened 2022-03-31
Make login page be landing page and update to not show nav bar


AshlynAdamson:
References user story #277 

We want the landing page of BEIMA to be the login page. That needs to be updated, as well as not showing the nav bar on the login page as well. See this resource https://www.bezkoder.com/react-jwt-auth/

Issue #369, opened 2022-03-31
Update pages to check for auth token


AshlynAdamson:
References user story #277 

We need to update our current pages to check for a logged in user, and if the user logs out, either manually or due to their token expiring. If a user is logged out, we need to redirect them to the login page. See this resource https://www.bezkoder.com/react-jwt-auth/

PR #440, opened 2022-04-10
Auth JS service

### LINKED ISSUE #367: Add auth service JS code
References user story #277 

We need some service functions that handle logging a user in, creating a new user, logging out, and getting the current user from the local storage. See this example here https://www.bezkoder.com/react-jwt-auth/

AshlynAdamson:
Closes #367

Getting auth service functions for the front end created. I used this example as I was developing, https://www.bezkoder.com/react-jwt-auth/. I followed the same idea for the page title stuff for the currentUser constant. There will most likely be changes needed here as more functionality gets integrated

### DIFF
--- a/BEIMA.Client/src/services/Authentication.js
+++ b/BEIMA.Client/src/services/Authentication.js
@@ -66,7 +66,7 @@ export default function logout(remember){
  * before the timeout occurs
  * 
  */
-export default function checkCurrentUser(){
+export default function checkAndSetCurrentUser(){
   const [setCurrentUser] = useOutletContext();
 
   let token = localStorage.getItem("currentUser");
--- a/BEIMA.Client/src/services/Authentication.js
+++ b/BEIMA.Client/src/services/Authentication.js
@@ -1,6 +1,5 @@
 import axios from 'axios';
 import jwt_decode from 'jwt-decode';
-import { useOutletContext } from 'react-router-dom';
 const API_URL = process.env.REACT_APP_API_URL;
 
 /**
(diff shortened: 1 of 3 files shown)


BEIMA.Client/src/services/Authentication.js
```diff
@@ -0,0 +1,89 @@
...
+  setCurrentUser({});
+}
+
+/**
+ * Checks if the token is in localStorage and sets the currentUser constant
+ * This is used for if a user selects remember me, but closes their browser
+ * before the timeout occurs
+ * 
+ */
+export default function checkCurrentUser(){
```
josephmoore99, line 69:
```suggestion
export default function checkAndSetCurrentUser(){
```

BEIMA.Client/src/services/Authentication.js
```diff
@@ -0,0 +1,89 @@
...
+
+      setCurrentUser(user);
+      return;
+    });
+  
+    return login.response;
+}
+
+/**
+ * Removed the token from the browser storage
```
josephmoore99, line 48:
```suggestion
 * Remove the token from the browser storage
```

josephmoore99:
Do we have functionality on all the pages to check that we are logged in? Or I guess how would the frontend respond if we tried to access a page where we did not have full permissions on?

AshlynAdamson:
> Do we have functionality on all the pages to check that we are logged in? Or I guess how would the frontend respond if we tried to access a page where we did not have full permissions on?

This will be done in tasks #369 and #370. If a user tries to access something they don't have access to, like the Users page, we can redirect them to the Devices page. Or if a user isn't logged in and tries to redirect using the url, it will redirect them to the login page

josephmoore99 (approved):

AshlynAdamson:
I am closing this PR since I ended up changing the work I did here in this PR, https://github.com/cs481-ekh/s22-beima/pull/441,  and I'll just keep it all in one PR

PR #441, opened 2022-04-10
Adding auth tokens to all JS service functions

### LINKED ISSUE #367: Add auth service JS code

### LINKED ISSUE #368: Update current JS service function with authentication bearer headers
References user story #277 

We need to update the current request service functions with the authentication header. See these resources, https://www.bezkoder.com/react-jwt-auth/, https://jwt.io/introduction

AshlynAdamson:
Closes #368 
Closes #367 

Adding auth headers to all the JS services. I will update the calls to these function in this task #369.

### DIFF
--- a/BEIMA.Client/src/services/AddBuilding.js
+++ b/BEIMA.Client/src/services/AddBuilding.js
@@ -1,4 +1,5 @@
 import axios from 'axios';
+import {getCurrentUser} from './Authentication.js';
 const API_URL = process.env.REACT_APP_API_URL;
 
 /**
@@ -7,9 +8,11 @@ const API_URL = process.env.REACT_APP_API_URL;
  * @param The user details to add to the DB
  * @return Error message or the inserted building ID
  */
-export default async function addBuilding(buildingDetails, token) {
+export default async function addBuilding(buildingDetails) {
+  let user = getCurrentUser();
+
   //performs the post and returns an error message or the inserted building ID
-  const dbCall = await axios.post(API_URL + "building", buildingDetails, {headers : {Authorization : `Bearer ${token}`}}).catch(function (error) {
+  const dbCall = await axios.post(API_URL + "building", buildingDetails, {headers : {Authorization : `Bearer ${user.token}`}}).catch(function (error) {
       if (error.response) {
         return error.response;
     }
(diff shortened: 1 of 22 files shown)


BEIMA.Client/src/services/AddBuilding.js
```diff
@@ -7,9 +7,9 @@ const API_URL = process.env.REACT_APP_API_URL;
  * @param The user details to add to the DB
  * @return Error message or the inserted building ID
  */
-export default async function addBuilding(buildingDetails) {
+export default async function addBuilding(buildingDetails, token) {
```
josephmoore99, line 10: This kinda feels bad having to pass the token in every time, plus I think this breaks all of the logic that calls these functions currently. Is there possibly a way for us to store this token in some sort of global variable that we set on login?

AshlynAdamson: Actually, it doesn't break it. I thought it would too, but it gets ignored by the backend since the endpoints don't check for it yet. 

In my other PR, I created a useState Constant in App.js to store the currentUser, which will hold the token. My idea was when we make these service calls, then we retrieve the token from the currentUser variable and pass it to the function. I could change it to use the useOutlet and grab the token from the currentUser variable

josephmoore99: I like that approach better, it removes the redundancy for us to have to remember to pass the token in every time we want to hit an endpoint.

AshlynAdamson: We may have to pass the token as a parameter. I updated them to use the useOutletContext, but since these functions aren't components, it doesn't like it. I will have to update my other auth service function as well 
image

josephmoore99: maybe we can define a function that returns the token that we can call?

AshlynAdamson: I think I can do that in App.js and put it in the AppLayout component. I will play around with it some more

AshlynAdamson:
- [ ] merge this first https://github.com/cs481-ekh/s22-beima/pull/440

AshlynAdamson:
> * [x]  merge this first [Auth JS service #440](https://github.com/cs481-ekh/s22-beima/pull/440)

This is no longer needed. I ended up changing the auth service while working with the header stuff

BEIMA.Client/src/services/AddBuilding.js
AshlynAdamson: Okay, I have changes up that use a function call to the auth service to get the token rather than passing it in as a parameter

josephmoore99 (approved):

PR #445, opened 2022-04-11
#217 Indicate More Tile Data

### LINKED ISSUE #217: Field tiles do not show all data/don't indicate there is more data
References user story #

If a fields input is long enough, then it will not be shown all the way in the input field. It is also not clear enough that a user can scroll to see the rest of the data. We should make the input field size change with the length of the input, or wrap the input, or indicate that there is more input to see.

**Steps to Reproduce:**
1. View an individual device with long field values, like serial number.
2. Notice that field tile with long value has value cut off.
3. Notice that there is no indication that there is more data to see.

**Actual Results:**

https://user-images.githubusercontent.com/58532471/157291336-571008af-6296-4aca-8b9f-3506eeaf479c.mp4

**Expected Results:**
Either wrap the data, make the input field size be dynamic, or indicate the there is more data to see.

Kenny-Miller:
Closes #217 

Add ability to expand field tile in order to see all of the text on the device and device type pages. When not expanded the input will show "fieldtext..." Additionally fix some css issues on the device and device type list view that would hide the details button when title of an item was very long.

image
image
image
image

### DIFF
--- a/BEIMA.Client/src/pages/DeviceTypes/DeviceTypePage.js
+++ b/BEIMA.Client/src/pages/DeviceTypes/DeviceTypePage.js
@@ -1,10 +1,11 @@
-import { useState, useEffect } from "react"
+import { useState, useEffect, useRef } from "react"
 import {ItemCard} from "../../shared/ItemCard/ItemCard"
 import * as Notifications from '../../shared/Notifications/Notification.js'
 import styles from './DeviceTypePage.module.css'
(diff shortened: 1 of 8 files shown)


AshlynAdamson (approved):
Looks good! I made task #446 (assigned to myself) to do the same on the add item pages since I think this same idea would be handy there as well.

PR #447, opened 2022-04-12
Auth integration into pages

### LINKED ISSUE #369: Update pages to check for auth token

### LINKED ISSUE #370: Make login page be landing page and update to not show nav bar

AshlynAdamson:
Closes #369 
Closes #370

Login page is now the landing page. Admin users only have access to the users page. A user can't delete themselves. And the remember me box works. I need to update the JS service functions to check for 401 response and log the user out if so, but I will do that in another task.

https://user-images.githubusercontent.com/58532471/163063050-04f799b3-6336-4288-a674-88bf2f504dbd.mp4

### DIFF
--- a/BEIMA.Client/src/App.js
+++ b/BEIMA.Client/src/App.js
@@ -33,7 +33,7 @@ const App = () => {
     <BrowserRouter>
       <Routes>
         <Route path="/" element={<AppLayout />}>
-          {currentUser.Token !== '' ?
+          {currentUser.token !== '' ?
           <>
           <Route index element={<Navigate to="/devices" />}/>
           <Route path="addDevice" element={<AddDevicePage/>}/>
@@ -72,7 +72,7 @@ const AppLayout = () => {
 
   return (
     <div className="page">
-      {currentUser.Token === '' ?
+      {currentUser.token === '' ?
       <></>
       : <NavBar />}
       <div className="content">
--- a/BEIMA.Client/src/App.js
+++ b/BEIMA.Client/src/App.js
@@ -33,7 +33,7 @@ const App = () => {
     <BrowserRouter>
       <Routes>
         <Route path="/" element={<AppLayout />}>
-          {currentUser.token !== '' ?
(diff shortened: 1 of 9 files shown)


BEIMA.Client/src/App.js
```diff
@@ -26,12 +27,15 @@ import PageTitle from './shared/PageTitle';
...
 
 const App = () => {
+  const [currentUser] = useState(getCurrentUser());
+
   return (
     <BrowserRouter>
       <Routes>
         <Route path="/" element={<AppLayout />}>
+          {currentUser.token !== '' ?
+          <>
```
josephmoore99, line 37: Interesting, didn't know this was a thing

BEIMA.Client/src/shared/NavBar.js
```diff
@@ -27,16 +32,22 @@ const NavBar = () => {
               <Row className="pageLink">
                 <Link to="/buildings" className="sharedText">Buildings</Link>
               </Row>
+              {currentUser.Role === 'admin' ?
```
josephmoore99, line 35: I'm wondering, should we only allow regular users to interact with just devices? Or do we still want to give them access to both buildings and device types, or one or the other?

josephmoore99 (approved):

thessbsu (approved):

BEIMA.Client/src/shared/NavBar.js
AshlynAdamson: I think access to everything besides Users is fine. Maybe a user could be entering a device that is in a new building and will need to add the building as well. Same thing for device type

PR #456, opened 2022-04-14
#414#413 filter model

### LINKED ISSUE #412: Update js service to use query string for the device lists call
references #411

### LINKED ISSUE #413: Add form/dropdowns to allow selection of filtering options
references #411 

Update devices page to have a form that allows users to select filtering parameters.

### LINKED ISSUE #414: Add current filter indicator to show selected filtering options
references #411

Kenny-Miller:
Closes #414
Closes #413 
Closes #412

Adds two dropdowns on the devices page in order to allow filtering of devices based on device type and building.

image

### DIFF
--- a/BEIMA.Client/src/pages/Devices/DevicesPage.js
+++ b/BEIMA.Client/src/pages/Devices/DevicesPage.js
@@ -1,35 +1,161 @@
-import { useEffect, useState } from "react"
+import { useEffect, useState, useMemo, useRef } from "react"
 import styles from  "./DevicesPage.module.css"
 import ItemList from "../../shared/ItemList/ItemList";
 import { useOutletContext } from 'react-router-dom';
+import { Button } from "react-bootstrap";
 import GetDeviceList from "../../services/GetDeviceList";
 import GetDeviceType from '../../services/GetDeviceType';
 import GetBuilding from "../../services/GetBuilding";
+import GetBuildingList from '../../services/GetBuildingList.js';
+import * as Notifications from '../../shared/Notifications/Notification.js';
+import GetDeviceTypeList from '../../services/GetDeviceTypeList.js';
+import * as Constants from '../../Constants.js';
+import Select from 'react-select';
+
+/*
+  * gets the list of buildings from the database
+  * @return the object containing building list
+  */
+const getBuildingList = async () => {
+  const buildingList = await GetBuildingList();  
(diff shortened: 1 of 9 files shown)


josephmoore99:
The filter items button feels a little out of place where it's at, I think it might look better if the two filter buttons were closer together?

Kenny-Miller:
> The filter items button feels a little out of place where it's at, I think it might look better if the two filter buttons were closer together?

Something like this? They will still expand horizontally when more selections are made.
image
image

BEIMA.Client/src/pages/Devices/DevicesPage.js
```diff
@@ -1,43 +1,169 @@
...
+              styles={customStyles}
+              ref={buildingSelect}
+            />
+          </div>
+          <Button id="clearFilterButton" onClick={() => {clearFilters(); filterCallback([],[])}}>Clear Filters</Button>
+        </div>
+        <Button id="submitFilterButton" 
+          disabled={loading} 
+          onClick={() => filterCallback(mapSelectedFilters(deviceTypeSelect), mapSelectedFilters(buildingSelect))}
+        >Filter Items</Button>
```
josephmoore99, line 124: I think it makes it more clear if we use "Apply Filters" here

josephmoore99:
> > The filter items button feels a little out of place where it's at, I think it might look better if the two filter buttons were closer together?
> 
> Something like this? They will still expand horizontally when more selections are made. image image

I was thinking more along the lines of having the two buttons next to each other at the end. But idk, UI is hard

Kenny-Miller:
> > > The filter items button feels a little out of place where it's at, I think it might look better if the two filter buttons were closer together?
> > 
> > 
> > Something like this? They will still expand horizontally when more selections are made. image image
> 
> I was thinking more along the lines of having the two buttons next to each other on the same row on the right. But idk, UI is hard, this does look a little better tho

Misread your initial comment. Something like this then? 
image

thessbsu:
That's good, but I would agree to swap the buttons

josephmoore99:
> That's good, but I would agree to swap the buttons

Agreed, also I would use "Apply Filters" instead of "Filter Items".

Kenny-Miller (commented):
Finally set of changes:
image

josephmoore99 (approved):

thessbsu (approved):

AshlynAdamson (approved):

PR #468, opened 2022-04-15
removed log statements

### LINKED ISSUE #450: remove console.log  statements from AddDevice.js
References user story #278

console.log statements for device image/docs need to be removed since they were present to allow deployment on Azure until they were used elsewhere.
They are now used in the save device function and should be ok to delete

thessbsu:
Closes #450

photo/document vars are no longer logged to the console on save


AshlynAdamson (approved):
Did a search in BEIMA.Client for console.log in vscode and only saw these lines

josephmoore99 (approved):

PR #471, opened 2022-04-16
#407 building endpoint auth

### LINKED ISSUE #407: Add authentication to all the building endpoints
References user story #16

Add checking for an authentication token before running any building endpoints. If the token is invalid or doesn't exist, throw a 401.
If token is valid continue on with the request

In addition update the last modified user value to use the username in the claims object that is gotten from the auth token.

Kenny-Miller:
Closes #407 

Adds authentication to the front of all the building functions as well as updates the last modified user to use the username in the claims token for the add and update building endpoint.

### DIFF
--- a/BEIMA.Backend.FT/BuildingFT.cs
+++ b/BEIMA.Backend.FT/BuildingFT.cs
@@ -79,6 +79,7 @@ namespace BEIMA.Backend.FT
             Assert.That(ex?.StatusCode, Is.EqualTo(HttpStatusCode.NotFound));
         }
 
+        [Test]
         public void NoBuildingsInDatabase_Unauthorized_BuildingGet_ReturnsUnauthorized()
         {
             var ex = Assert.ThrowsAsync<BeimaException>(async () =>
@@ -89,6 +90,7 @@ namespace BEIMA.Backend.FT
             Assert.That(ex?.StatusCode, Is.EqualTo(HttpStatusCode.Unauthorized));
         }
 
+        [Test]
         public void NoBuildingsInDatabase_Unauthorized_BuildingGet_EmptyId_ReturnsNotFound()
         {
             var ex = Assert.ThrowsAsync<BeimaException>(async () =>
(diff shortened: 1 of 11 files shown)


BEIMA.Backend.FT/BuildingFT.cs
```diff
@@ -66,6 +79,25 @@ await TestClient.GetBuilding(id)
...
+        {
+            var ex = Assert.ThrowsAsync<BeimaException>(async () =>
+                await UnauthorizedTestClient.GetBuilding("1234567890abcdef12345678")
+            );
+            Assert.IsNotNull(ex);
+            Assert.That(ex?.Message, Does.Contain("Invalid credentials."));
+            Assert.That(ex?.StatusCode, Is.EqualTo(HttpStatusCode.Unauthorized));
+        }
+
+        public void NoBuildingsInDatabase_Unauthorized_BuildingGet_EmptyId_ReturnsNotFound()
```
josephmoore99, line 92: We already have a test case for this on NoBuildingsInDatabase_BuildingGet_ReturnsNotFound

josephmoore99:
We should add unit tests specifically testing the unauthorized response, as I think we are doing this for all the other endpoints as well.

BEIMA.Backend.FT/BuildingFT.cs
Kenny-Miller: That test case doesn't test if the client is unauthorized though.

josephmoore99: That's true, but this one still seems to be a not found error over an unauthorized error. But I suppose it shouldn't hurt to leave it in.

BEIMA.Backend.FT/BuildingFT.cs
```diff
@@ -66,6 +79,25 @@ await TestClient.GetBuilding(id)
             Assert.That(ex?.StatusCode, Is.EqualTo(HttpStatusCode.NotFound));
         }
 
+        public void NoBuildingsInDatabase_Unauthorized_BuildingGet_ReturnsUnauthorized(string id)
```
josephmoore99, line 82: This test is not being run because there is no [Test] tag.

josephmoore99 (approved):

thessbsu (approved):

roguedeleter (approved):

AshlynAdamson (approved):

PR #502, opened 2022-04-19
Updating cypress tests to pass

### LINKED ISSUE #442: Update Cypress tests to start on login page, enter credentials, and then test
References user story #278 

After we finish the authentication functionality, we will need to update the cypress tests to start on the login page, enter in test credentials, and then continue testing.

AshlynAdamson:
Closes #442

image

Few notes for the technical documents

- there needs to be an admin user in the test database with the username 'testuser'. The tester will need to set the password in the test code for it to login correctly
- each item view page (device, device type, building, user) needs to have the id variable updated with an id from the database. The test should update this before running tests
- the add user page adds a user, but if the test is ran again, it will fail since we don't allow for duplicate users. this test added user will need to be deleted from the database


BEIMA.Client/src/pages/Building/BuildingPage.js
```diff
@@ -130,8 +130,8 @@ const BuildingPage = () => {
 
     const cancel = () => {      
       setNumber(building.number)
-      setLat(building.latitude)
-      setLong(building.longitude)
+      setLat(building.location.latitude)
+      setLong(building.location.longitude)
```
AshlynAdamson, line 134: These tests actually caught a bug lol

BEIMA.Client/cypress/integration/LoginPage.spec.js
```diff
@@ -11,24 +13,8 @@ describe('Login Page', () => {
...
-    })
-    cy.get('[id=submitBtn]').click()
-    cy.get('.invalid-feedback').each((val) => {
-      cy.wrap(val).should('be.visible')
-    })
-    cy.wait(1250)
-    cy.get('[id=submitBtn]').should('be.disabled')
-  })
-  
-
```
AshlynAdamson, line 29: I think due to the submit form change, we don't get the error feedback anymore, so this test fails. Instead, we get this, which I think is still fine. It also shows if the username is entered but no password is and the user tries to submit
image

BEIMA.Client/cypress/integration/ABeginningTest.spec.js
```diff
@@ -0,0 +1,18 @@
...
+describe("Login to BEIMA", () => {
+  it('Login to BEIMA', () => {
+    cy.visit('http://localhost:3000/')
+    cy.get('[id=username]').scrollIntoView().clear().type("testuser")
+
+    // REPLACE PASSWORD WITH VALID PASSWORD FROM TEST USER IN DATABASE
+    cy.get('[id=password]').scrollIntoView().clear().type("")
+    cy.get('[id=submitBtn]').click()
+
+    cy.wait(700);
```
josephmoore99, line 16: Is there a more reliable way for us to wait that is not just waiting for a specific amount of time for things to load.

BEIMA.Client/cypress/integration/AddDevicePage.spec.js
```diff
@@ -41,12 +42,12 @@ describe("Verify Data in fields is still present when invalid coords exist and A
     cy.visit('http://localhost:3000/addDevice')
     cy.get("#typeDropDown").scrollIntoView().click()
     cy.get("#typeDropDown").children().last().scrollIntoView().click()
-    cy.get('#inputBuilding').scrollIntoView().type("Student Union Building")
```
josephmoore99, line 44: Do we need to replace the testing functionality for buildings?

BEIMA.Client/cypress/integration/ABeginningTest.spec.js
```diff
@@ -0,0 +1,18 @@
+/// <reference types="cypress" />
+
+/// The purpose of this test is to login to BEIMA before all the other tests are ran
+/// to ensure that navigating through BEIMA can still happen.
+/// The test database must have a user with this username, but the tester needs to replace the empty
+/// password with a valid one.
+describe("Login to BEIMA", () => {
+  it('Login to BEIMA', () => {
+    cy.visit('http://localhost:3000/')
+    cy.get('[id=username]').scrollIntoView().clear().type("testuser")
```
josephmoore99, line 10: We could use our default user generated by the mongo connector here instead?

BEIMA.Client/cypress/integration/ABeginningTest.spec.js
AshlynAdamson: We can wait for an item to be visible and have a timeout for it, I will play around with that

BEIMA.Client/cypress/integration/ABeginningTest.spec.js
AshlynAdamson: We could, but that would assume the developer running this has no users in the database. And if that default login gets changed then this would need to be changed as well

BEIMA.Client/cypress/integration/AddDevicePage.spec.js
AshlynAdamson: The building and device type dropdown use the same shared component, and we have tests for the device type dropdown, so I think we are okay without replacing the building input tests

AshlynAdamson:
- [ ] look into beforeEach and afterEach

BEIMA.Client/cypress/integration/ABeginningTest.spec.js
AshlynAdamson: Now that this is in the before block, I think waiting is a bit easier, and when running them in the GUI it still does it really fast,  I'm not sure if it waits the whole 7 seconds is what it feels like

josephmoore99 (approved):

roguedeleter (approved):

PR #517, opened 2022-04-21
420 all devices report building name

### LINKED ISSUE #420: Update reports to include the building name instead of building Id
References user story #278

Currently the report service produce reports that contain a building id instead of the building's name. Change it to show the building name.

Kenny-Miller:
Closes #420

Updated report writer to use building name instead of building id. Also uses the device type name instead of device type id.


AshlynAdamson (dismissed):

thessbsu (approved):

josephmoore99:
Do we have a task for device type id as well?

Kenny-Miller:
> Do we have a task for device type id as well?

We now do

AshlynAdamson (approved):

josephmoore99 (approved):

PR #525, opened 2022-04-21
first attempt at getting relative paths to work

### LINKED ISSUE #417: Updated build settings to run the application from a subdirectory and not root.
References user story #278
Since the sdp server will not allow us to run our frontend from root, we must make sure that it can be ran from a sub directory.
ex instead of https://sdp.../devices it should be https://sdp.../beima/devices

Should be tested by deploying and testing on the sdp server.

thessbsu:
Closes #417

Updated my local package.json with a "homepage" parameter.
The simple " . " should fix it so everything is served relative to the directory in which the index file is in, this may or may not work but there's other solutions to try

The only thing that was actually added is the homepage var, but for some reason my package.json was practically empty locally and I copy/pasted the contents from the main branch

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "PUBLIC_URL=/beima react-scripts build",
     "test": "react-scripts test",
     "cypress:test": "cypress run -b chrome && cypress run -b firefox && cypress run -b edge",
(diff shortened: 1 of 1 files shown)


AshlynAdamson (approved):

Kenny-Miller (approved):

thessbsu:
abandoning branch and PR, the weirdness with the package.json not matching doesn't sit well

PR #541, opened 2022-04-24
added host.json change

### LINKED ISSUE #540: Add host.json entry for API calls
References user story #434 

The addition to the host.json file was not included in the deploy fix PR

thessbsu:
Closes #540 

Added host.json change. Tested working without the "-service" on the end of the api calls over the network

### DIFF
--- a/BEIMA.Backend/host.json
+++ b/BEIMA.Backend/host.json
@@ -7,5 +7,10 @@
                 "excludedTypes": "Request"
             }
         }
+    },
+    "extensions": {
+        "http": {
+          "routePrefix": "beima"
+        }
     }
 }
--- a/BEIMA.Backend.FT/BeimaClient.cs
+++ b/BEIMA.Backend.FT/BeimaClient.cs
@@ -194,7 +194,7 @@ namespace BEIMA.Backend.FT
         /// <returns>The device with the given id.</returns>
         public async Task<Device> GetDevice(string id)
         {
-            var response = await SendRequest($"api/device/{id}", HttpVerb.GET);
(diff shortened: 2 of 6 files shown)


thessbsu:
not sure why tests are failing

thessbsu:
added proxy items and commented the creation in docker-compose. Future teams may need to test deploy

BEIMA.Backend/host.json
```diff
@@ -7,5 +7,10 @@
                 "excludedTypes": "Request"
             }
         }
+    },
+    "extensions": {
+        "http": {
+          "routePrefix": "beima"
```
AshlynAdamson, line 13: Could the functional tests be failing because of this added prefix?

josephmoore99: Yeah, I think you need to add the correct url prefix in the FT BeimaClient.cs, and the FT base, I can branch from this and do that real quick if needed?

thessbsu: You'll likely get it done faster than I will

Kenny-Miller (commented):
I think we will want a proxy added for the backed as well
to tell it requests on /beima-service/ should be directed to the backend

thessbsu:
> I think we will want a proxy added for the backed as well to tell it requests on /beima-service/ should be directed to the backend

while testing I removed the "-service" part and it worked with just the redirect on all "beima" paths

Kenny-Miller (approved):

josephmoore99 (approved):

AshlynAdamson (approved):

PR #543, opened 2022-04-24
Added beima prefix to FT

### LINKED ISSUE #542: Switch backend FT to use beima prefix
References user story #434

We need to switch over the backend FT to use the beima prefix instead of the api prefix.

josephmoore99:
Closes #542

Adds the beima url prefix to the backend FT

### DIFF
--- a/BEIMA.Backend.FT/BeimaClient.cs
+++ b/BEIMA.Backend.FT/BeimaClient.cs
@@ -194,7 +194,7 @@ namespace BEIMA.Backend.FT
         /// <returns>The device with the given id.</returns>
         public async Task<Device> GetDevice(string id)
         {
-            var response = await SendRequest($"api/device/{id}", HttpVerb.GET);
+            var response = await SendRequest($"device/{id}", HttpVerb.GET);
             var content = await response.Content.ReadAsStringAsync();
             return JsonConvert.DeserializeObject<Device>(content);
         }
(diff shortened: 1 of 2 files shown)


BEIMA.Backend.FT/BeimaClient.cs
```diff
@@ -194,7 +194,7 @@ public async Task<HttpResponseMessage> SendMultiPartRequest(string route, Object
         /// <returns>The device with the given id.</returns>
         public async Task<Device> GetDevice(string id)
         {
-            var response = await SendRequest($"api/device/{id}", HttpVerb.GET);
+            var response = await SendRequest($"device/{id}", HttpVerb.GET);
             var content = await response.Content.ReadAsStringAsync();
```
AshlynAdamson, line 198: I'm not sure the api part was removed, but just the beima part was added

josephmoore99: When I run it on my machine it seems to be replaced? 
image

AshlynAdamson: Ah okay, good to know

Kenny-Miller (approved):

BEIMA.Backend.FT/BeimaClient.cs
josephmoore99: Looks like I just needed to add in an extra slash was the problem

PR #544, opened 2022-04-25
Bug report writer

Kenny-Miller:
Fixes a bug where downloading a zip file from the backend running on a linux box was causing the zip to be unopenable without using 7zip. This had to do with the file names inside the zip. Also fixed an issue where commas and quotes were not properly being escaped which resulted in ill formatted rows.

### DIFF
--- a/BEIMA.Backend/ReportService/ReportWriter.cs
+++ b/BEIMA.Backend/ReportService/ReportWriter.cs
@@ -1,8 +1,11 @@
 ﻿using BEIMA.Backend.Models;
 using BEIMA.Backend.MongoService;
+using CsvHelper;
+using CsvHelper.Configuration;
 using MongoDB.Bson;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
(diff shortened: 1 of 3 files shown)


BEIMA.Backend/ReportService/ReportWriter.cs
```diff
@@ -350,7 +355,6 @@ private static List<string> GenerateDeviceReportHeaders(DeviceType deviceType)
         /// </summary>
         /// <param name="device">Device that's data values should be used</param>
         /// <param name="deviceType">DeviceType that the field keys should be based on</param>
-        /// <param name="buildings">Buildings that the building id should be based on</param>
```
josephmoore99, line 353: Why remove this?

Kenny-Miller: Not sure how that got removed. Will get it back in.

josephmoore99 (approved):

AshlynAdamson (approved):

roguedeleter (approved):

